PR #7, opened 2015-07-14
Stored is the MySQL counterpart of PERSISTENT

madhuracj:
Stored is the MySQL counterpart of PERSISTENT. See http://mysqlserverteam.com/generated-columns-in-mysql-5-7-5/


udan11:
Thanks!

PR #27, opened 2016-01-13
Re add php 5.4 for 3.0.* release

fezfez:
There is no compatibility issue with php 5.4, as the original 3.0.0 was compatible with php 5.4, i propose to  keep 3.0.* compatible with php 5.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 php:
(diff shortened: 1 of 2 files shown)


fezfez:
Any news ?

udan11:
Seems ok.

PR #28, opened 2016-01-19
Fixed 500 error when DESC syntax is used in query

elrossco22:
Fix for 500 error caused when the statement starts with DESC with is an abbreviation for DESCRIBE


udan11:
Thanks!

elrossco22:
@udan11 No problem, how do I go about getting this pulling this change into phpmyadmin as it originated from here (https://github.com/phpmyadmin/phpmyadmin/issues/11867)

also do you have any other issues on this project I could have a look at

PR #29, opened 2016-02-16
Fix index type is not recognized by parser, Issue#11971

007durgesh219:
Signed-off-by: Durgesh <007durgesh219@gmail.com>


udan11:
Looks good, thank you!

PR #30, opened 2016-02-17
Change return value from getAll in case of not parsed query

nijel:
Fixes https://github.com/phpmyadmin/phpmyadmin/issues/11609

Signed-off-by: Michal Čihař <michal@cihar.com>


udan11:
Thanks!

PR #35, opened 2016-02-24
Use motranslator for translating

nijel:
This makes it work transparently with or without phpMyAdmin.

See also https://github.com/phpmyadmin/phpmyadmin/pull/12023

Signed-off-by: Michal Čihař michal@cihar.com


src/Translator.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        return $this->translator->gettext($msgid);
+    }
+
+    /**
+     * Returns the singleton Translator object
+     *
+     * @return Translator object
+     */
+    public static function getInstance()
```
nisargjhaveri: Can't we make `gettext` a static method and use `Translator::gettext($msg)` instead of `Translator::getInstance()->gettext($msg)`. Looks much cleaner and easier to use.

nijel: Indeed that would be doable as well with lazy initialization of the translator (we don't need it in most cases, it's used only for errors).

nijel:
Thinking more about this I'm not sure if it's good idea to include this in releases used in phpMyAdmin 4.6.x as this will bring two gettext implementation into phpMyAdmin codebase....

nisargjhaveri:
If we want to think really 'modular', it shouldn't matter :P . But yes, makes sense.

How are we actually managing releases of SQL parser?

nijel:
It won't matter for 4.7, but for 4.6 I'm not sure if switching phpmyadmin to motranslator is reasonable...

nisargjhaveri:
Agreed.

nijel:
Superseded by https://github.com/phpmyadmin/sql-parser/pull/106

Issue #36, opened 2016-02-26
DROP VIEW statement is not constructed properly by the parser


007durgesh219:
var_dump of parser for following query ```DROP VIEW abc```

![dropview](https://cloud.githubusercontent.com/assets/7701929/13352631/83150bb6-dcb4-11e5-8d30-d304ce233107.png)

Fix is required for phpmyadmin/phpmyadmin#11966

nijel:
Fixed by https://github.com/phpmyadmin/sql-parser/pull/37

PR #37, opened 2016-02-26
Fix DROP VIEW statement is not constructed properly by the parser, Issue #36

007durgesh219:
Signed-off-by: Durgesh <007durgesh219@gmail.com>


nijel:
Thanks!

PR #63, opened 2016-08-18
fixup of example in documentation

shadiakiki1986:
Hi. I added a specific example query (select statement) and showed how to change the table in the select from "a" to "b"


nijel:
Merged, thanks!

PR #66, opened 2016-09-03
Fix #10: Add a check if we find a new start to new statement before delimiter

### LINKED ISSUE #10: ignore missed semicolons
test sql:
```sql
START TRANSACTION

ALTER TABLE `tbl` CHANGE `uid` `uid` INT( 10 ) UNSIGNED NOT NULL AUTO_INCREMENT

ALTER TABLE `tbl` CHANGE `field_1` `field_2` INT( 10 ) UNSIGNED NOT NULL

COMMIT
```

After each line we missed semicolons, but sql-parser don't see any errors.
Test code same as in #9

devenbansod:
Add a check if we find a new start to new statement before delimiter


nijel:
The code looks good, but again I'm missing testcase here....

PR #68, opened 2016-09-05
Fix #49: Add Support for 'CREATE TABLE `table_copy` LIKE `table`;'

### LINKED ISSUE #49: CREATE TABLE `table_copy` LIKE `table`;
Did not support creating copy of table by:
`CREATE TABLE 'table_copy' LIKE 'table';`

Output:
`An opening bracket was expected. (near: LIKE)` (after LIKE)
`At least one column definition was expected. (near:  )` (before LIKE)

devenbansod:
Fix #49 
Added Support for statements like 'CREATE TABLE \`table_copy\` LIKE \`table\`;'


nijel:
Please add testcase to cover this query as well...

nijel:
Still some cases are still not covered:

* Missing LIKE expression
* Building the query

nijel:
Thanks! Sorry for being so nitpicking, but I really thing all such things need test coverage so that we see when we break it by other changes in future.

PR #69, opened 2016-09-07
Fix spaces after expression in PartitionDefinition's build

devenbansod:
Fixes phpmyadmin/phpmyadmin#12545


nijel:
Looks good. Maybe we should include testcase for the query from phpmyadmin/phpmyadmin#12545 as well?

PR #70, opened 2016-09-07
Fix #59: Non-reserved keywords should be allowed as a field name

### LINKED ISSUE #59: Non-reserved keywords should be allowed as a field name
Did not support set 'action' when creating a table prior to create a trigger:
"action varchar(50) default null"

Output:
A symbol name was expected! (near action)

### LINKED ISSUE #72: Query with keyword field name built wrong
```php
$parser = new SqlParser\Parser('INSERT INTO `table_name` (`order`) VALUES (1);');
$parser->parse();

echo $parser->statements[0]->build();
// INSERT  INTO `table_name`(order) VALUES (1)
```

The original query becomes invalid and can not be executed in MySQL (missed &#96; around field name &#96;order&#96;).

devenbansod:
While trying to parse a fieldname, disallow only if it is a reserved keywords.
Reserved keywords are still allowed if backquotes are used.

All column names in Insert statement would be backquoted by default.

Added testcases to cover both with and w/o backquotes scenarios.

Fix #59 
Fix #72 


nijel:
Merged and added missing test case in 4afb228.

PR #76, opened 2016-09-12
Fix parsing of INSERT...SELECT and INSERT...SET syntax

### LINKED ISSUE #50: Unrecognized keyword ON, DUPLICATE and KEY
### Steps to reproduce
```
INSERT INTO `champs` (`id`,`key`) VALUES (412,'Thresh')
ON DUPLICATE KEY UPDATE id=412,key='Thresh'
```

### Actual behaviour
```
#1: Unrecognized keyword. (near "ON" at position xx)
#2: Unrecognized keyword. (near "DUPLICATE" at position xx)
#3: Unrecognized keyword. (near "KEY" at position xx)
```

### Server configuration
**Web server:**
Apache/2.4.17 (Win32) OpenSSL/1.0.2d PHP/5.6.21

**Database:**
Database client version: libmysql - mysqlnd 5.0.11-dev - 20120503 - $Id: 76b08b24596e12d4553bd41fc93cccd5bac2fe7a $

**phpMyAdmin version:**
4.6.1

## Same Issue?
[Unrecognized keyword FULL and OUTER #12205](https://github.com/phpmyadmin/phpmyadmin/issues/12205)

I am not sure if the SQL Command are 100% right, but i hope so.
Would be nice if someone could confirm this bug.

### LINKED ISSUE #73: INSERT ... SELECT Syntax
SqlParser doesn't support [INSERT ... SELECT Syntax](http://dev.mysql.com/doc/refman/5.7/en/insert-select.html).

```php
$parser = new SqlParser\Parser('INSERT INTO `a` (`value`) SELECT `b`.`value` FROM `b`;');
$parser->parse();

echo $parser->statements[0]->build();
// Notice: Trying to get property of non-object in .../Components/ArrayObj.php on line 191
// Warning: implode(): Invalid arguments passed in .../Components/ArrayObj.php on line 191
// INSERT  INTO `a`(value) VALUES ()
```

devenbansod:
Implement separate parse function for InsertStatement

Fix #50
Fix #73
Fix phpmyadmin/phpmyadmin#11628


nijel:
Good, only test coverage could be better ;-). (see https://codecov.io/gh/phpmyadmin/sql-parser/compare/bd8936d2d6d1b932211a29cea313749be13bac83...7e6fbf70c33e0320fee040f596aa788cb3135805)

nijel:
Added two more tests:

* 98c2a72 Add test for building INSERT ... SELECT ...
* c7cc3ed Add test for building INSERT ... SET.

While trying to add third test, I've noticed that the ON DUPLICATE handling is still not working as expected, see #78.

PR #77, opened 2016-09-12
Fix parsing of REPLACE INTO ... Statements

### LINKED ISSUE #48: REPLACE INTO ... SELECT ... not implemented
Hello,

I am trying to lint [some SQL files](https://git.wikimedia.org/blob/labs%2Ftools%2Fheritage.git/master/erfgoedbot%2Fsql%2Ffill_table_monuments_all.sql) using [cweiske/php-sqllint](https://github.com/cweiske/php-sqllint) which uses sql-parser under the hood.

I bumped into an issue with a SELECT inside a REPLACE INTO ([third form](https://dev.mysql.com/doc/refman/5.5/en/replace.html)). I tested it on the latest version of sql-parser
Here is an example (shamelessly taken from RandomStackOverflowQuestionExample)
```
./bin/lint-query --query "REPLACE INTO sales( 'item_id', 'date', 'qty', 'price' ) SELECT item_id, date, qty, price FROM sales_to_accept"
```
fails with `#1: A new statement was found, but no delimiter between it and the previous one. (near "SELECT" at position 56)`

Hope that helps!

devenbansod:
Fix parsing of REPLACE INTO ... Statements.

Now supports all three variations with REPLACE INTO .. VALUES, REPLACE INTO .. SET, REPLACE INTO .. SELECT.

Fix #48


nijel:
Can you please add tests for query building? It's currently not covered at all...

nijel:
Thanks!

PR #79, opened 2016-09-13
Stop SelectStatement parsing and revert back in case of ON DUPLICATE KEY  …

### LINKED ISSUE #78: INSERT ... SELECT ... ON DUPLICATE KEY UPDATE broken
While ON DUPLICATE KEY UPDATE works for regular INSERT statements, it does not for INSERT ... SELECT as the SELECT parser seems to consume the ON DUPLICATE part:

```
./bin/lint-query --query 'INSERT  INTO tbl SELECT  * FROM bar ON DUPLICATE KEY UPDATE baz = 1'
#1: Unrecognized keyword. (near "KEY" at position 49)
#2: A new statement was found, but no delimiter between it and the previous one. (near "UPDATE" at position 53)
#3: Unexpected token. (near "UPDATE" at position 53)
#4: Unexpected token. (near "=" at position 64)
#5: Unexpected token. (near "1" at position 66)
```

devenbansod:
Fix parsing of INSERT INTO ... SELECT .... ON DUPLICATE KEY

Fix #78


nijel:
Thanks!

PR #82, opened 2016-09-13
fix default options in Formater constructor

remicollet:
In the construction, default value for 'line_ending' doesn't work, as $this->options['type']  cannot be tested at this point (not yet set).

So for "html", default value still "\n", as shown in the test suite.
This fix should give the expected (documented) behavior.


nijel:
Good catch, thanks!

PR #85, opened 2016-09-20
Fix parsing of user@host without backquotes

devenbansod:
Introduce an option to parse 'expr=' (like var, var=, expr)

Fix tests

Fix phpmyadmin/phpmyadmin#12298


nijel:
Looks good!

nijel:
I've added changelog entry in 00eb348

PR #94, opened 2016-10-21
Enable Lexing of begin_label in the in-procedure statements

### LINKED ISSUE #57: inapt error symbol when labeling a loop in a routine
### Problem definition
When creating a routine using SQL commands in console, if we use label for loop 
(eg: student_loop : LOOP), it identifies colon (:) as an "unexpected token" and shows error symbol at that line.
1_shows_as_an_error

But when we execute (i.e: when pressing "Go" button), it works fine without pushing any errors.
2_works_fine
 
Then when try to export the routine, again it shows as there is an error at the line which defines the label, but which is actually not an error.
3_export_of_routine

### Expected behaviour
When creating a label for a loop, colon following the label should not be detected as an error.

### Actual behaviour
An error symbol occurs at the line which defines label for a loop; but which is actually not an error. 
Also when executing the query, it works fine.

### Server configuration

**Web server:** xampp server 

**Database:** MySQL

**phpMyAdmin version:** 4.6.2

### Client configuration
**Browser:** Google chrome

**Operating system:** Windows 8.1

devenbansod:
Enable Lexing of begin_label in the in-procedure statements
For Ex. BEGIN.. END, LOOP, REPEAT, WHILE statements

Fix #57
Fix phpmyadmin/phpmyadmin#12344
Fix phpmyadmin/phpmyadmin#12228


nijel:
Thanks!

PR #95, opened 2016-10-24
Fix #93 : Fix parsing of NATURAL JOIN, CROSS JOIN and related joins

### LINKED ISSUE #93: "Natural" not recognized
As I was researching http://stackoverflow.com/questions/39374782/sql-query-with-limit-doesnt-work-in-phpmyadmin, I discovered that the 'natural' keyword is not recognized as part of a JOIN query:

```
SELECT  * FROM `myTable` 
  natural JOIN `column1` 
  natural JOIN `column2` 
WHERE `mytable`.`something` = 1 
ORDER BY mytable_date DESC LIMIT 0,4
```

Or, more simply (and actually tested on 'sakila' from the CLI client):
```
select * from actor natural join actor_info;
```

devenbansod:
Fix parsing of NATURAL JOIN, CROSS JOIN and related joins
Fix broken parsing of multiple JOINS in a single query.

Fix #93 

Add tests to cover parsing of these joins.


nijel:
Thanks, looks great!

PR #96, opened 2016-11-03
Fix #22 : Validate order of clauses in the parsing of statements

### LINKED ISSUE #22: switched WHERE and LIMIT not detected
I do not get an error when syntax checking a query with switched LIMIT and WHERE clauses:

```
SELECT pid, name2 FROM tablename LIMIT 10 WHERE pid = 20
```

devenbansod:
Validate order of clauses in the parsing of statements

Fix #22


nijel:
Merged with minor fixes, thanks!

PR #99, opened 2016-11-12
Add parsing End options like FOR UPDATE and LOCK IN SHARE MODE in Select

devenbansod:
Add parsing End options like FOR UPDATE and LOCK IN SHARE MODE in Select

Add tests for parsing and building.


nijel:
Thanks!

PR #100, opened 2016-11-15
Fix broken incorrect clause order detection for Joins

devenbansod:
Fixing #22 by #96 lead to this breaking. It gives incorrect messages even if the query and the clauses are in right order.

This PR should remove that incorrect message.


devenbansod:
Hi @nijel I am unable to comprehend the error in the travis build and its actual source from the commit  ? Plus it fails only on hhvm.

nijel:
The hhvm builds fail since some hhvm update, I haven't yet found time to look at it...

PR #101, opened 2016-11-21
Fix incorrect error in union queries

devenbansod:
Prevent incorrect error messages in UNION queries (which were introduced in #96 )


nijel:
Thanks

PR #104, opened 2016-12-10
Minor PHPDoc fix

uuf6429:
I've noticed that both when the property is written to or read from, an array of ArrayObj is expected.


nijel:

PR #111, opened 2016-12-30
Handling merging formats arrays

bigfoot90:
Currently if you want to change a format you have to pass a complete formats array in the options.
By this PR you can now pass a partial array of formats, eg.:
* change only (one or many) formats identified by a specific type or flag
* override only `html` or ony `cli` or only `function` of selected formats

Written tests to prevent downgrading

### DIFF
--- a/src/Utils/Formatter.php
+++ b/src/Utils/Formatter.php
@@ -55,128 +55,201 @@ class Formatter
      */
     public function __construct(array $options = array())
     {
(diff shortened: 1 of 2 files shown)


bigfoot90:
What are the failing tests?

nijel:

PS: The checks on coverage and code quality are there to help us spot problematic code, in your case it was rather caused by bug in PHP coverage...

PR #112, opened 2017-01-03
Apply php-cs-fixer

bigfoot90:
Apply php-cs-fixer
https://github.com/FriendsOfPHP/PHP-CS-Fixer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change Log
 
+## [Unreleased]
+
 ## [3.4.15] - 2017-01-02
 
 * Fix return value of Formatter.toString() when type is text


.gitignore
```diff
@@ -1,3 +1,5 @@
+.php_cs.cache
+cs_fixer_tmp_*
```
ibennetch: I'm not terribly inclined to add these to .gitignore, but I'll wait to see what the others think about it.

src/Component.php
```diff
@@ -7,9 +7,8 @@
  * There is a small difference between *Component and *Keyword classes: usually,
  * *Component parsers can be reused in multiple  situations and *Keyword parsers
  * count on the *Component classes to do their job.
- *
- * @package SqlParser
```
ibennetch: I have to admit I'm a little stumped by this, I've looked at several styleguides and none of them mention this. The phpdoc documentation still mentions it, so it doesn't seem to be deprecated. I'm just confused about it.

src/Component.php
```diff
@@ -19,21 +18,20 @@
...
 {
-
     /**
      * Parses the tokens contained in the given list in the context of the given
      * parser.
      *
-     * @param Parser     $parser  The parser that serves as context.
-     * @param TokensList $list    The list of tokens that are being parsed.
-     * @param array      $options Parameters for parsing.
+     * @param Parser     $parser  the parser that serves as context
```
ibennetch: This also doesn't quite make sense, and conflicts with the samples provided [here](https://www.phpdoc.org/docs/latest/getting-started/your-first-set-of-documentation.html#what-does-a-docblock-look-like). For me, it's much more readable to have the sentence (or phrase) start with a capital letter. The trailing period, I'm rather indifferent about; ideally I don't mind seeing it go away since these are often only phrases anyway.

src/Components/AlterOperation.php
```diff
@@ -17,86 +15,84 @@
...
      */
     public static $DB_OPTIONS = array(
-        'CHARACTER SET'             => array(1, 'var'),
-        'CHARSET'                   => array(1, 'var'),
-        'DEFAULT CHARACTER SET'     => array(1, 'var'),
-        'DEFAULT CHARSET'           => array(1, 'var'),
-        'UPGRADE'                   => array(1, 'var'),
-        'COLLATE'                   => array(2, 'var'),
-        'DEFAULT COLLATE'           => array(2, 'var'),
+        'CHARACTER SET' => array(1, 'var'),
```
ibennetch: I tend to prefer having everything lined up, but I'm not strictly tied to that.

src/Components/AlterOperation.php
```diff
@@ -17,86 +15,84 @@
...
+        'PARTITIONING' => 2,
+        'PRIMARY KEY' => 2,
+        'SPATIAL' => 2,
+        'TABLESPACE' => 2,
+        'INDEX' => 2,
     );
 
     /**
-     * All view options
+     * All view options.
```
ibennetch: Now the tool seems to be making things up; earlier it was removing periods after full sentences and now it's adding them to a few words. I'm not sure what to make of this.

src/Components/AlterOperation.php
```diff
@@ -121,20 +117,20 @@ class AlterOperation extends Component
...
     public static function parse(Parser $parser, TokensList $list, array $options = array())
     {
-        $ret = new AlterOperation();
+        $ret = new self();
 
         /**
          * Counts brackets.
          *
-         * @var int $brackets
+         * @var int
```
ibennetch: This seems wrong to me; I don't think it should be removing the variable name from the docblock.

src/Components/AlterOperation.php
```diff
@@ -248,18 +244,19 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
-     * @param array          $options   Parameters for building.
+     * @param AlterOperation $component the component to be built
+     * @param array          $options   parameters for building
      *
      * @return string
      */
     public static function build($component, array $options = array())
     {
-        $ret = $component->options . ' ';
+        $ret = $component->options.' ';
```
ibennetch: Looking at some comments on PSR-1 (not the standard itself, which doesn't seem to say anything), this seems opposite the intentions of the PSR. In all the discussion and examples around it, they use the spaces on both sides. I think the spaces make it much more readable, so I'd vote to keep the spaces, too.

src/Components/AlterOperation.php
```diff
@@ -121,20 +117,20 @@ class AlterOperation extends Component
...
+     * @param Parser     $parser  the parser that serves as context
+     * @param TokensList $list    the list of tokens that are being parsed
+     * @param array      $options parameters for parsing
      *
      * @return AlterOperation
      */
     public static function parse(Parser $parser, TokensList $list, array $options = array())
     {
-        $ret = new AlterOperation();
+        $ret = new self();
```
ibennetch: This is an interesting change.

src/Components/CaseExpression.php
```diff
@@ -1,106 +1,98 @@
...
-     * @param TokensList $list    The list of tokens that are being parsed.
+     * @param Parser     $parser the parser that serves as context
+     * @param TokensList $list   the list of tokens that are being parsed
      *
      * @return Expression
      */
     public static function parse(Parser $parser, TokensList $list, array $options = array())
     {
-        $ret = new CaseExpression();
+        $ret = new self();
```
ibennetch: This is also interesting.

ibennetch:
Hi @bigfoot90 and thanks for your submission. Some of the changes definitely look good, and some of them are not things I personally would change. I'd like to see some discussion from the rest of the team on these.

For phpMyAdmin itself, we use [our own hybrid (and loose) styleguide](https://github.com/phpmyadmin/phpmyadmin/wiki/Developer_guidelines#coding-style) which mostly based on the PEAR styles. It might be good to enhance this and particularly to update it (for instance, PSR now makes recommendations about namespaces that weren't possible a few years ago).

src/Component.php
nijel: I think it makes less sense since namespaces were introduced, so it's probably okay to remove.

src/Component.php
nijel: IMHO it looks better in the documentation as sentences, but I'm quite okay with both choices.

class_sqlparser_component_-_2017-01-03_10 08 33

src/Components/AlterOperation.php
nijel: On the other side the lined up parts are harder to maintain (you need to reindent whole block when adding something longer).

src/Components/AlterOperation.php
nijel: Probably it's a sentence for it as it starts with capital :-)

src/Components/AlterOperation.php
nijel: I also prefer to have spaces around docs.

nijel:
Anyway I think we should first agree on coding style and then enforce it, not the other way around :-).

However I find this really useful, I've added it to motranslator and shapefile libs (with minor adjustments such as whitespace around `.` operator) right now as the outcome is less problematic there:

* https://github.com/phpmyadmin/motranslator/commit/cf71d7cabec1b3197e8c376025ea7c16566d7cb1 (configuration)
* https://github.com/phpmyadmin/motranslator/commit/d29162c455862e9b5d0021c6593f870c06c3f282 (fixes)
* https://github.com/phpmyadmin/shapefile/commit/ec6191f28939c6b1780f748b517a10c01ee696c2 (configuration)
* https://github.com/phpmyadmin/shapefile/commit/43b86b7782562d2d8dcd9315fed4b9ab7053de7c (fixes)

.gitignore
bigfoot90: I've installed `php-cs-fixer` as global vendor, normally I take this configuration in my global .gitignore,
but I thought for the other people who can run the tool, they can not have this configuration.


I've seen also, that you have .idea in your .gitignore, this line should be put in your global .gitignore configuration since .idea isn't realted to the project but to the ide used by programmer which can be different from person to person.

src/Components/AlterOperation.php
```diff
@@ -17,86 +15,84 @@
  * Parses an alter operation.
```
bigfoot90: Duplicated comment, see before `namespace`

src/Component.php
bigfoot90: This behavior is configurable

src/Components/AlterOperation.php
bigfoot90: PhpDoc makes different `summary` and `description` sections. https://www.phpdoc.org/docs/latest/guides/docblocks.html#summary
**Summay** *The summary is a short but effective overview of an element.*
**Description** *A description can be a long text with an elaborate explanation what the associated element does. The description is optional.*

As I read, the texts after `@var` and `@param` are descriptions
https://www.phpdoc.org/docs/latest/references/phpdoc/tags/var.html#examples

src/Components/AlterOperation.php
bigfoot90: The name is required only if you refer
https://www.phpdoc.org/docs/latest/references/phpdoc/tags/var.html

>In this case the name of the property MAY be omitted

src/Components/AlterOperation.php
bigfoot90: This behavior is configurable

src/Components/CaseExpression.php
bigfoot90: This will do the same

bigfoot90:
Sorry, closed this  wrong.

PR #117, opened 2017-01-06
Some clean, remove useless brackets

bigfoot90:
Another small fix:
- Remove useless brackets from Fromatter
- Make more readable ContextGenerator


nijel:

PR #118, opened 2017-01-06
composer autoload-dev

bigfoot90:
Use composer's `autoload-dev`


nijel:

PR #119, opened 2017-01-06
Make the vendor package smaller

bigfoot90:
Ignore files when exporting to make the vendor smaller


nijel:

PR #120, opened 2017-01-06
Better Formatter tests

bigfoot90:
Writing tests this way can be harden, but you can be sure that your feature/fix for an formatter type will not impact on any other formatter type.

WDYT?
If you think that can be useful I'll continue to write tests for the remaining tests cases.

### DIFF
--- a/src/Utils/Formatter.php
+++ b/src/Utils/Formatter.php
@@ -350,8 +350,6 @@ class Formatter
 
             // Checking if pointers were initialized.
             if ($prev !== null) {
-                /** @var Token $prev */
-
                 // Checking if a new clause started.
                 if (static::isClause($prev) !== false) {
                     $lastClause = $prev->value;
--- a/src/Utils/Formatter.php
+++ b/src/Utils/Formatter.php
@@ -322,7 +322,7 @@ class Formatter
         /**
          * Previously parsed token.
          *
-         * @var Token
+         * @var Token|null
          */
         $prev = null;
 
@@ -349,12 +349,9 @@ class Formatter
             }
 
             // Checking if pointers were initialized.
-            /**
-             * Previous Token.
-             *
-             * @var Token $prev
-             */
             if ($prev !== null) {
+                /** @var Token $prev */
+
                 // Checking if a new clause started.
                 if (static::isClause($prev) !== false) {
                     $lastClause = $prev->value;
--- a/src/Utils/Formatter.php
+++ b/src/Utils/Formatter.php
@@ -349,8 +349,10 @@ class Formatter
             }
 
             // Checking if pointers were initialized.
-            /*
+            /**
              * Previous Token.
+             *
+             * @var Token $prev
              */
             if ($prev !== null) {
                 // Checking if a new clause started.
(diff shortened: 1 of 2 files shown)


nijel:
I think this is good idea, just make sure that all things currently covered will stay covered.

tests/Utils/FormatterTest.php
```diff
@@ -230,6 +230,58 @@ public function mergeFormats()
...
+                    '&nbsp;&nbsp;&nbsp;&nbsp;*',
+            ),
+            'select with comment' => array(
+                'query' => 'select 1 # Comment',
+                'text' =>
+                    "SELECT\n" .
+                    "    1 # Comment" . "\n",
+                'cli' =>
+                    "\x1b[35mSELECT" . "\n" .
+                    "    \x1b[92m1 \x1b[37m# Comment\\x0A" . "\x1b[0m",
```
bigfoot90: Found first issue writing tests.
Why there is a `\x0A` in the comment?

bigfoot90: Fixing PR https://github.com/phpmyadmin/sql-parser/issues/123

nijel:
Any progress on this?

bigfoot90:
Yes, I was waiting for #123 to rebase onto.

bigfoot90:
I'm ready for the review, yo can see that every test case in `testFormat` is covered by the new tests in `testFormat_new`.
Any comments or suggestions?

Actually only the `codacy/pr` tests are failing, this will be resolved when I remove the old tests in `testFormat` and replace `testFormat_new` with `testFormat`.

src/Utils/Formatter.php
```diff
@@ -349,8 +349,10 @@ public function formatList($list)
             }
 
             // Checking if pointers were initialized.
-            /*
+            /**
              * Previous Token.
+             *
+             * @var Token $prev
```
nijel: This is really not variable declaration, so the comment does not make much sense here...

src/Utils/Formatter.php
```diff
@@ -349,10 +349,9 @@ public function formatList($list)
             }
 
             // Checking if pointers were initialized.
-            /*
-             * Previous Token.
-             */
             if ($prev !== null) {
+                /** @var Token $prev */
+
```
bigfoot90: @nijel **WDYT** about this way?
I known that there is not a variable declaration, this way I say a bit different thig.
Here tha annotation just says that there is a variable `$prev` of type `Token`.

This annotation is demanded by IDE, else it says that there is not any property `value` on `null`.

If it is not ok I can revert this one, no problem.

nijel: But $prev is declared before, I see no reason to have it annotated here...

bigfoot90: There is already a annotation on the declaration *(on line 325)*, but PhpStorm is simply ignoring it.

bigfoot90:
@nijel Removed `@var` anotation.
Are there other issues with this PR?
I can proceed to fix the method name `testFormat_new` and remove the old deprecated tests?

nijel:
I think it's in good shape, please go ahead.

nijel:
Thanks, I've merged it.

The coverage failure was caused by formatter handling differently `BEGIN` and `begin` as you've changed this. I've fixed this in 5fdfa5b0767c8b62e5dd16f1b99465c2d59c9d2f and 1a1168169c3d3acdf7ee3c5f1a839bcb66829297 and adjusted tests accordingly in 65fb611e091ef7d0b28039126733848bd20edb9d.

bigfoot90:
Just to be clear. This way I'm testing that keywords are recognized and uppercased correctly.

nijel:
They were uppercased properly, but the formatting didn't recognize because of being lowercase...

PR #125, opened 2017-01-21
Update .gitattributes

remicollet:
For consistency, as tests are back.


nijel:
Thanks, same fix applied to motranslator and shapefile libs.

PR #130, opened 2017-01-24
Replace \x1b with \e

bigfoot90:
**Do not merge yet**

My IDE suggest me to replace `\x1b` characters with `\e`.
I've done the replace only in tests to check if some thing breaks.
Is this a more optimized way to write the same character?

### DIFF
--- a/tests/Utils/CLITest.php
+++ b/tests/Utils/CLITest.php
@@ -29,12 +29,12 @@ class CLITest extends TestCase
         return array(
             array(
                 array('q' => 'SELECT 1'),
-                "\x1b[35mSELECT\n    \x1b[92m1\x1b[0m\n",
+                "\e[35mSELECT\n    \e[92m1\e[0m\n",
                 0,
             ),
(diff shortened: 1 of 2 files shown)


bigfoot90:
Reading the web I see that `\x1b` is equal to `\e` and to `\0x1B` and `\033`. 
That's ever the same character.

I prefer `\e` because it's more readable.

But what about these lines https://github.com/phpmyadmin/sql-parser/pull/130/files#diff-5897b34c7e97d0c27240a808838bc33eR371?
Where `text\e[33m` will become `text\\x1B[33m`

If it's ok for you I can continue with this PR.

bigfoot90:
**:-(**   I see now the results from travis, it seems that php5.3 does not support parsing of `\e` characters.
Closing.

nijel:
That's exactly reason why we don't use it, see 3ece9eed.

bigfoot90:
OK, good to known.

Issue #129, opened 2017-01-23
Dependency of phpmyadmin/motranslator


bigfoot90:
As from version `4.1.0` https://github.com/phpmyadmin/sql-parser/commit/438ac0bd5751cfc7582c03768d0149404004bb15#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R20
You are requesting `phpmyadmin/motranslator` as a dependency.

I'm using SqlParser in my standalone project (outside of `phpmyadmin`) mainly as a formatter tool

**My question is:**
Is this really a hard depencedependencyncy?
I think that's a dependency of `phpmyadmin/phpmyadmin` itself.
Can we [suggest installing](https://getcomposer.org/doc/04-schema.md#suggest) `motranslator` instead of requiring it?

nijel:
Patches are welcome to make the dependency optional. It's used to localize error messages from the parser.

nijel:
Also note there is one problem with suggests in composer - AFAIK there is no way to define which version you need, so you can easily end up with installed version with incompatible API.

bigfoot90:
Right, composer does not have version requirements for suggestions, but have `conflict` where you can exclude the incompatible versions, in addition we can add a note in the `README.md`.

nijel:
Okay, that would work as well.

The only part which needs to be adjusted is https://github.com/phpmyadmin/sql-parser/blob/master/src/Translator.php

bigfoot90:
Ok, I can prepare a PR in the next days.

Can you show me how to use this `Translator` in a real life example? Please.

nijel:
It just returns translated text, for example see https://github.com/phpmyadmin/sql-parser/blob/438ac0bd5751cfc7582c03768d0149404004bb15/src/Component.php#L43

PR #132, opened 2017-02-01
Made motranslator vendor optional

bigfoot90:
Suggest installing `phpmyadmin/mostranslator` instead of requiring it.

Close https://github.com/phpmyadmin/sql-parser/issues/129


nijel:

PR #134, opened 2017-02-07
[CS] Other cs fixes  proposal

bigfoot90:
**WDYT**?

* Sort use imports
* Remove useless `else`
* Remove useless `return null`
* Add phpdoc `@param` annotation where missing

Something to add/change/remove?

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -13,7 +13,11 @@ return PhpCsFixer\Config::create()
         '@Symfony' => true,
         'array_syntax' => array('syntax' => 'long'),
         'concat_space' => array('spacing' => true),
+        'ordered_imports' => true,
         'no_trailing_whitespace' => true,
+        'no_useless_else' => true,
+        'no_useless_return' => true,
+        'phpdoc_add_missing_param_annotation' => true,
         'phpdoc_order' => true,
     ))
     ->setFinder($finder)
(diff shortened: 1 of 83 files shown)


src/Components/CreateDefinition.php
```diff
@@ -226,11 +226,10 @@ public static function parse(Parser $parser, TokensList $list, array $options =
                         );
 
                         return $ret;
-                    } else {
+                    }
                         // Non-reserved keywords are allowed without backquotes
                         $expr->name = $token->value;
```
nijel: It failed to adjust indentation here.

.php_cs
```diff
@@ -13,7 +13,11 @@ return PhpCsFixer\Config::create()
         '@Symfony' => true,
         'array_syntax' => array('syntax' => 'long'),
         'concat_space' => array('spacing' => true),
+        'ordered_imports' => true,
         'no_trailing_whitespace' => true,
+        'no_useless_else' => true,
+        'no_useless_return' => true,
+        'phpdoc_add_missing_param_annotation' => true,
```
nijel: I think it's better to make at least separate commit, but most likely also PR for each of the changes. The `ordered_imports` certainly looks okay, I'm not so sure about `no_useless_*`.

bigfoot90:
This was replaced by #136 #137 #138 #139

PR #135, opened 2017-02-12
Apply cs fixer

bigfoot90:
This is a base PR for #136 #137 #138 #139

PR #137, opened 2017-02-12
Cs fix phpdoc add missing param annotation

bigfoot90:
Extracted from #134 
Based on #135 

Add missing `param` annotation


bigfoot90:
Is there a way to exclude `tests/` directory from this fix?
I'm not entirely convinced.

PR #141, opened 2017-02-16
Inline JOINs

bigfoot90:
**Before**:
```sql
JOIN
    tbl2
ON
    c1 = c2
JOIN
    tbl3
ON
    c3 = c4
```

**After**:
```sql
JOIN tbl2 ON c1 = c2
JOIN tbl3 ON c3 = c4
```


nijel:

PR #142, opened 2017-02-17
Fix broken clause order validation

### LINKED ISSUE #113: Broken code for clause order validation
The current code is broken, at least the part looking at JOIN clauses - it contains unreachable code (I've spotted this when looking at code not covered by tests).

Code is here: https://github.com/phpmyadmin/sql-parser/blob/master/src/Statement.php#L480

```php
            // Handle ordering of Multiple Joins in a query
            if ($clauseStartIdx != -1) {
                if ($joinStart == 0 && stripos($clauseType, 'JOIN') !== false) {
                    $joinStart = 1;
                } elseif ($joinStart == 1 && stripos($clauseType, 'JOIN') === false) {
                    $joinStart = 2;
                } elseif ($joinStart == 2 && stripos($clauseType, 'JOIN') !== false) {
                    $error = 1;
                }
            }
```

The problem is that all JOIN clauses are in same group in the `$CLAUSES` array, so the error condition can never happen as there will be no `$clauseType` containing `JOIN` after `$joinStart` is set to 2.

devenbansod:
Fix #113


nijel:
Thanks!

PR #143, opened 2017-02-18
Parse LOAD statement properly

### LINKED ISSUE #131: Handle LOAD queries
Currently the parser does not properly handle LOAD queries. This causes problems in phpMyAdmin as it does not recognize the LOAD query as the one to display number of affected rows, see https://github.com/phpmyadmin/phpmyadmin/issues/12345.

The minimal support should cover properly setting query flags for LOAD, so at least is_affected and is_insert should be set (I didn't look at others, this needs further look).

The MySQL documentation: https://dev.mysql.com/doc/refman/5.7/en/load-data.html

devenbansod:
Fix #131

Might help to fix phpmyadmin/phpmyadmin#12345


nijel:
Thanks!

PR #146, opened 2017-03-31
Fixes striping of inline comments #11717

Achilles-96:
Signed-off-by: Raghuram Vadapalli<raghuram.vadapalli@research.iiit.ac.in>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ composer.lock
 coverage.xml
 *~
 .php_cs.cache
+*sw[op]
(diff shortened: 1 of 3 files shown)


nijel:
I'm not really sure here as one of the purposes of BufferedQuery was to strip comments. Maybe @udan11 or @devenbansod can share their opinion as well.

Currently BufferedQuery does strip all comments (except for MySQL commented statements). This was probably done for performance reasons. However this removes them from bodies stored on the server (procedures, triggers, ...), what is not really desired.

I see two options:

1. We stop stripping comments completely. This is what this patch does.
2. We parse the query and keep the comments where necessary.

I think second approach will perform even worse than keeping all comments, so I think this is way to go.

nijel:
Thanks, merged!

PR #148, opened 2017-04-04
Fixed inline comments parsing

nijel:
This is different approach than #147 to fix https://github.com/phpmyadmin/phpmyadmin/issues/13139.

Please review the changes.


src/Lexer.php
```diff
@@ -546,6 +546,10 @@ public function parseComment()
             ) {
                 $token .= $this->str[$this->last];
             }
+            // Include trailing \n as whitespace token
+            if ($this->last < $this->len) {
+                $this->last--;
+            }
```
devenbansod: I am not very familiar with the Lexer code, but I believe this makes sense. Having the whitespaces as a separate whitespace token makes sense.

PR #149, opened 2017-04-11
Fixes wrong extract of string tokens with escaped characters.

### LINKED ISSUE #40: Wrong extraction of string with escaped quote
Try parsing `SELECT '\''`. The extracted value is `\`, while it should be `'`. The bug is in `Token::extract` for `Token::TYPE_STRING`, it does just blindly replace `''` with single one, what is not right solution in this case.

udan11:
Fixes #40.

Signed-off-by: Dan Ungureanu <udan1107@gmail.com>


nijel:
Thanks!

PR #153, opened 2017-05-16
Don't print duplicated cli formatting characters

bigfoot90:


nijel:
Merged, thanks (I've initially tried to fix conflict in the GitHub, but I've messed it up, so I've merged the original commit manually).

Issue #154, opened 2017-05-28
Error using GROUP BY with PostgreSQL


JoseNetoooo:
Hello everybody! I've been using sql-parser with postgresql. When I build the query and get the rawSql I see the following:
```sql
GROUP BY "created_at" ASC
```
...but PostgresSQL does not accept **ASC** after the **GROUP BY** keyword. Is there a way to avoid it? Or Is there anything I'm missing here?

Above we can see the build() function located in the Components/OrderKeyword.php class. This class is used by the **GROUP BY** keyword and always returns:

```php
$component->expr . ' ' . $component->type
```

```php
public static function build($component, array $options = array())
    {
        if (is_array($component)) {
            return implode(', ', $component);
        }

        return $component->expr . ' ' . $component->type;
    }
```

JoseNetoooo:
[https://develdocs.phpmyadmin.net/sql-parser/source-class-PhpMyAdmin.SqlParser.Parser.html#349-552](The Parser class)

[https://develdocs.phpmyadmin.net/sql-parser/source-class-PhpMyAdmin.SqlParser.Components.OrderKeyword.html#14-140](The OrderKeyword class)

nijel:
Neither MySQL does accept ASC/DESC here, it's bug in the parser that uses another component here...

nijel:
Fixed by https://github.com/phpmyadmin/sql-parser/pull/155 with few other fixes:

* 9bdebcf Document change for #154 and #155
* 4af06d2 Remove not used type arg from GROUP BY
* f18666a Merge commit '48d61a0c05316a25d9462379e79ff56d5e22a29c'
* 9837c18 Add (failing) test for building GROUP BY clause

williamdes:
Using MariaDB 10.2 the query was valid https://github.com/phpmyadmin/sql-parser/commit/4af06d24b041e499fb0e75ab3a98caf9a91700ef#commitcomment-36623272

PR #155, opened 2017-05-31
Add GroupKeyword class to fix postgres GROUP BY

JoseNetoooo:
As related here [https://github.com/phpmyadmin/sql-parser/issues/154](url) by me, PostgresSQL does not accept **ASC** or **DESC** after the **GROUP BY** keyword. My purpose is to create a new class called GroupKeyword with an empty default type. Why to create a new class? I think because maybe we need to be more specific in some scenarios.

Anyway, thank you so much for this Sql Parser.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "phppgadmin/sql-parser",
+    "name": "jtfnetoo/sql-parser",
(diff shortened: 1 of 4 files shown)


nijel:
Thanks, I've merged the functional changes. Quite obviously the composer changes are not for merge...

JoseNetoooo:
Ok, Thanks! ...and sorry for the composer changes in that pull request!

PR #173, opened 2017-09-12
Fix #172 Support parameter binding

### LINKED ISSUE #172: Support named parameters
phpMyAdmin support parameter binding with named parameters in SQL editor.
For example, you can write a query like 
``
INSERT INTO `person` (`firstname`, `lastname`, `email`) VALUES (:firstname, :lastname, :email);
``
and phpMyAdmin will allow specifying values for the parameters.

However, the parser does not support this and triggers syntax error warnings on the editor. It would be nice to have the sql-parser support this feature.

madhuracj:


nijel:
Looking at testsuite failures, this breaks handling of assignment operator`:=`.

madhuracj:
@nijel Thanks for pointing that out. I think all is well now.

nijel:
Thanks, indeed looks good now.

PR #176, opened 2017-10-10
Update CreateStatement.php

Lapkritinis:
I think there is mistake in code as well:

                    if ($brackets === 0) {
                        $this->$field = trim($this->$field);
                        $field = null;
                    }

### DIFF
--- a/src/Statements/CreateStatement.php
+++ b/src/Statements/CreateStatement.php
@@ -515,7 +515,7 @@ class CreateStatement extends Statement
                     }
 
                     // Building the expression used for partitioning.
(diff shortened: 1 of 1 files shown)


src/Statements/CreateStatement.php
```diff
@@ -515,7 +515,7 @@ public function parse(Parser $parser, TokensList $list)
                     }
 
                     // Building the expression used for partitioning.
-                    $this->$field .= ($token->type === Token::TYPE_WHITESPACE) ? ' ' : $token->token;
+                    $field .= ($token->type === Token::TYPE_WHITESPACE) ? ' ' : $token->token;
```
nijel: This code is definitely correct.

PR #183, opened 2017-12-15
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertArrayHasKey` and `assertArrayHasKey` instead of `isset` function;
- `assertNull` instead of comparison with `null` keyword.


nijel:

Issue #192, opened 2018-05-16
Syntax error on keyword 'as' with case in a valid query 


sveencoop:
### Steps to reproduce

The query 
``` 
SELECT
c.nome, 
case c.id_estado
    when 1 then 12 
    when 2 then 27
    when 3 then 13 
    when 4  then 16 
    when 5 then 29
    when 6 then 23
    when 7 then 53 
    when 8 then 32 
    when 9 then 52 
    when 10 then 21
    when 11 then 31
    when 12 then 50 
    when 13 then 51
    when 14 then 15 
    when 15 then 25 
    when 16 then 26 
    when 17 then 22 
    when 18 then 41 
    when 19 then 33 
    when 20 then 24 
    when 21 then 11 
    when 22 then 14 
    when 23 then 43 
    when 24 then 42 
    when 25 then 28 
    when 26 then 35 
    when 27 then 17 
end as teste,
c.cod_municipio, cll.latitude, cll.longitude
FROM cidade c
INNER JOIN  cidades_lat_long cll ON c.cod_municipio = cll.id_geocode
```

The Structure

```
CREATE TABLE cidades_lat_long (
    id_geocode int(11) NOT NULL,
    latitude double NOT NULL,
    longitude double NOT NULL,
    PRIMARY KEY (id_geocode)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;

CREATE TABLE cidade (
    id int(11) NOT NULL AUTO_INCREMENT,
    nome varchar(45) NOT NULL,
    id_estado int(11) NOT NULL,
    cod_municipio int(11) DEFAULT NULL COMMENT 'CEP',
    PRIMARY KEY (`id`),
    KEY fk_cidade_estado(id_estado)
) ENGINE=InnoDB AUTO_INCREMENT=10036 DEFAULT CHARSET=utf8;
```

![capturar](https://user-images.githubusercontent.com/10080622/40099119-0ae04c58-58b5-11e8-9e7c-69697d39bdad.PNG)

### Server configuration
**Operating system**:
Windows 10

**Web server:**
Localhost (with Apache and MYSQL, installed by WAMP)

**PHP version:**
5.6.31
**phpMyAdmin version:**
4.7.4

### Client configuration
**Browser:**
Opera 52.0

**Operating system:**
Windows 10

sinri:
I met this issue as well, and if you add brackets outside the `CASE...END` it would run correctly.
[...]

PR #193, opened 2018-05-30
Sinri

sinri:
Try to fix https://github.com/phpmyadmin/sql-parser/issues/192
By automatically add brackets around the `CASE...END` clause.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ coverage.xml
 *~
 .php_cs.cache
 *sw[op]
+debug/
+composer.phar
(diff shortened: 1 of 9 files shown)


ibennetch:
Hi @sinri, what's the status of this pull request? It looks like there are still some comments and debugging code throughout to me.

sinri:
@ibennetch Sorry for late.
I met that issue in developing my own project and found in issue list months ago. I have no more time to wait for the official update so I had to fork and write my own solution. I wondered whether I could do the perfect modification on this but I thought that I might share my code to you.
I checked just now, that issue is still open so that you might be working on it. If you need this idea to fix, just use it.

Issue #162, opened 2017-07-31
Aliases not supported after CASE expression


nijel:
Aliases do not work after CASE expression:

```
$ ./bin/lint-query --query 'select 1 name'
#1: Unrecognized keyword. (near "name" at position 9)
$ ./bin/lint-query --query 'select case  when 1 > 0 then "x" else "y" end AS name'
#1: Unrecognized keyword. (near "AS" at position 46)
#2: Unrecognized keyword. (near "name" at position 49)
```

See also #161 and https://github.com/phpmyadmin/phpmyadmin/issues/13541

ibennetch:
Thank you for reporting. This has been fixed with #203 and will be part of the 4.3.0 release of sql-parser.

ibennetch:
I'm going to leave this opened because while the second query works, the first one (`./bin/lint-query --query 'select 1 name'`) still gives an error.

ibennetch:
@mostertb Perhaps you have some thought about the first query as well, since you did a great job on the other parts. Not a huge deal if you don't have time or interest, but since you were just working with that part of the code it may still be fresh in your mind.

PR #203, opened 2018-10-13
Add support for Aliases on CASE expressions

mostertb:
Hi 

This PR allows CASE expressions to be aliased to fix issues https://github.com/phpmyadmin/sql-parser/issues/192 and https://github.com/phpmyadmin/sql-parser/issues/162

It specifically adds support for `CASE ... END [AS] alias` for both *CASE* formats with the `AS` keyword being optional.

Im a little out of my depth so any help or guidance tests or functionality that I have missed would be great.

I know that queries like `CASE ... END expression [AS] alias` such as `SELECT CASE WHEN 2 IS NULL THEN 0 ELSE 2 END + 1 as 'foo'` are not covered. Let me know if there is sufficient use-case for this scenario and I'll attempt it in a new PR


ibennetch:

Issue #204, opened 2018-11-19
CREATE OR REPLACE VIEW build()s incorrectly


travisghansen:
```
CREATE OR REPLACE VIEW
```
becomes
```
CREATE OR REPLACE=VIEW
```

devenbansod:
@travisghansen I tried to add tests for building the CREATE OR REPLACE VIEW at #209  and they seem to pass with the expected output. 

Am I missing something?

devenbansod:
@travisghansen closing this as it seems to be working as expected.

Please re-open if you find something misbehaving.

travisghansen:
@devenbansod I updated `v4.2.4 => v4.2.5` and this appears to be resolved now.  So good for me.

PR #209, opened 2018-12-20
Add tests for builder of CREATE or REPLACE VIEW statement

devenbansod:
Issue #204 seems to point a breakage in building of Create or Replace View. I did not find it, but thought if I could add a test, I would provide it as a reference to close the issue.

### DIFF
--- a/tests/Builder/CreateStatementTest.php
+++ b/tests/Builder/CreateStatementTest.php
@@ -271,6 +271,18 @@ EOT
             'SELECT id, first_name FROM employee WHERE id = 1 ',
             $stmt->build()
         );
+
+        $parser = new Parser(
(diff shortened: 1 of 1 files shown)


devenbansod:
@williamdes thanks for the review.

At https://github.com/phpmyadmin/sql-parser/issues/204, it was reported that the building of the CREATE or REPLACE VIEW statement is broken, but I found it to be working correctly.

I decided to add this test so that I could demonstrate to the issue reporter that it indeed is working as expected.

So, with this PR (since the CI is green), we could close #204 saying it's working as expected (and requesting the author to re-open the issue if the library does misbehave while building some specific variant of CREATE or REPLACE VIEW)

PR #220, opened 2019-01-08
Bump minimum version to PHP 7.1

mauriciofauth:
Drop support for PHP 5.3, PHP 5.4, PHP 5.5, PHP 5.6, PHP 7.0 and HHVM

Since phpMyAdmin requires PHP 7.1, it makes sense to require PHP 7.1 for sql-parser as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,33 +3,24 @@ language: php
 dist: trusty
 
 php:
-  - "5.4"
-  - "5.5"
-  - "5.6"
-  - "7.0"
-  - "7.1"
-  - "7.2"
-  - "hhvm-3.12"
-  - "hhvm-3.18"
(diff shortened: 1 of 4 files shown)


ibennetch:
Anyone who still uses the version 4 can pull it from the Git tag v4.3.1 or GitHub release 4.3.1, minimizing the importance of having the branch — but I think that creating the branch now will encourage fixes to be backported to v4 by authors, so I do think it's a good idea and will take care of it.

ibennetch:
We should decide on a name for the branch, though; I'm partial to `v4` or `version4`...keeping with the phpMyAdmin style we might also prefer `QA_4_3`. Any opinions?

mauriciofauth:
@ibennetch What about `4.x`? I also like `v4`.

Issue #221, opened 2019-01-16
Parser cannot deal truncate statement


sinri:
(1) First, cannot parse common truncate statement.

```php
$query="truncate a";
$parser = new Parser($query,true);
```

would throw 

```
PHP Fatal error:  Uncaught PhpMyAdmin\SqlParser\Exceptions\ParserException: Unexpected token. in /Users/sinri/code/phpstorm/sql-parser/src/Parser.php:617
Stack trace:
#0 /Users/sinri/code/phpstorm/sql-parser/src/Statement.php(247): PhpMyAdmin\SqlParser\Parser->error('Unexpected toke...', Object(PhpMyAdmin\SqlParser\Token))
#1 /Users/sinri/code/phpstorm/sql-parser/src/Statement.php(90): PhpMyAdmin\SqlParser\Statement->parse(Object(PhpMyAdmin\SqlParser\Parser), Object(PhpMyAdmin\SqlParser\TokensList))
#2 /Users/sinri/code/phpstorm/sql-parser/src/Parser.php(516): PhpMyAdmin\SqlParser\Statement->__construct(Object(PhpMyAdmin\SqlParser\Parser), Object(PhpMyAdmin\SqlParser\TokensList))
#3 /Users/sinri/code/phpstorm/sql-parser/src/Parser.php(381): PhpMyAdmin\SqlParser\Parser->parse()
#4 /Users/sinri/code/phpstorm/sql-parser/doc/1.php(14): PhpMyAdmin\SqlParser\Parser->__construct('truncate a', true)
#5 {main}
  thrown in /Users/sinri/code/phpstorm/sql-parser/src/Parser.php on line 617
```

Only `TRUNCATE table x.y` is accepted.

(2) Cannot build an existed TruncateStatement.

Which would return empty string.

Version Commit 28427543

sinri:
Tried to parse two kinds of keyword in one query string.

`truncate table `a`.`b`;select 1`

got

```
PhpMyAdmin\SqlParser\Lexer::lex@230 METHOD: parseKeyword MET TOKEN {"token":"truncate","value":"truncate","keyword":"TRUNCATE","type":1,"flags":33,"position":null}
PhpMyAdmin\SqlParser\Lexer::lex@230 METHOD: parseWhitespace MET TOKEN {"token":" ","value":" ","keyword":null,"type":3,"flags":0,"position":null}
[...]

PR #222, opened 2019-01-16
Fix issue with parsing TRUNCATE statements, Issue #221

sinri:
Make the following SQL parsable, which is mentioned in issue #221 

`truncate table `a`.`b`;truncate c;select truncate from truncate;`

### DIFF
--- a/src/Statement.php
+++ b/src/Statement.php
@@ -320,11 +320,15 @@ abstract class Statement
             }
 
             // Checking if this is the beginning of a clause.
-            if (! empty(Parser::$KEYWORD_PARSERS[$token->value]) && $list->idx < $list->count) {
-                $class = Parser::$KEYWORD_PARSERS[$token->value]['class'];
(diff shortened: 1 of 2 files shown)


sinri:
Wonder why conflicts came out.

sinri:
Merged the latest master branch to resolve conflicts.

src/Token.php
```diff
@@ -343,4 +343,14 @@ public function getInlineToken()
             $this->token
         );
     }
+
+    /**
+     * Make you happier when you need to debug.
```
ibennetch: Is this function valuable to include in the code or is it debugging code that should be removed?

If it's worth keeping, I think the comment here should be changed to make the purpose of the function more clear.

williamdes:
Merged as 04da59b51db477a9caaf3e755df414cf1b3c09c3
Thanks for your contribution and patience @sinri !

sinri:
Great Result Finally!

Issue #227, opened 2019-02-15
Builder produces SQL code with redundant spaces


pszalko:
Statement builder adds redundant spaces in produced SQL code, for example:
`SELECT 1;` becomes `SELECT  1`.

pszalko:
Fixed by PR #228

devenbansod:
Fixed by #228, closing.

PR #228, opened 2019-02-15
Fix redundant whitespaces - Issue #227

pszalko:
Changed implementation of `build()` methods to generate more consistent SQL.

This commit fixes implementation of `build()` methods to produce more consistent SQL code without redundant spaces.


williamdes:
@devenbansod You can merge this one :)

PR #232, opened 2019-03-11
Repair simulated delete query syntax error

dcolter:
When a DELETE query was simulated, the result would be a MySQL syntax error, because the WHERE clause had ORDER BY ... added to it and the ORDER BY clause had LIMIT ... added to it.  

The actual query (OK button) finished normally

See the discussion in [Simulated Delete Query with ORDER BY and LIMIT Causes Error](https://github.com/phpmyadmin/phpmyadmin/issues/14995)

### DIFF
--- a/src/Utils/Query.php
+++ b/src/Utils/Query.php
@@ -648,7 +648,7 @@ class Query
                 }
             }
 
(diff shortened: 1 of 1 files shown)


devenbansod:
Closing this, since the original issue has been fixed in #233

Issue #237, opened 2019-03-30
add support to lint a *.sql file


staabm:
would be great to have support for linting sql files.

that way we would be able to use this tool for linting purposes within a CI pipeline

staabm:
Additionally/Alternatively it would be great to lint the stdin

ibennetch:
Excellent idea.

This was implemented with #239.

This will be part of the next `sql-parser` release. I wonder if this is worth a major version release, jumping to version 5, or if we do a minor version increase to version 4.4. Any opinions from anyone?

staabm:
Thank you guys. From my POV would a minor bump be sufficient (semver wise) - in case there are no other breaking changes in the pipeline

staabm:
I just realized that in the changelog is mentioned that a lot of old PHP version will be dropped of support.

Therefore a new major version is a must

ibennetch:
Oh, right, I was planning that anyway and had just forgotten for a moment.
We're looking to merge a few other feature-based pull requests before
pushing the new release, if we can get that done soon.

Thanks again!

staabm:
Any chance we get a new release soon?

PR #238, opened 2019-04-02
Support --path option in highlight query

staabm:
this PR adds a `--path` option for the highlight command.
refs https://github.com/phpmyadmin/sql-parser/issues/237

this is meant as a simplified initial proposal so I can get early feedback.
after we agreed on how this option should look like in the highlight-case, I will adjust the remaing 2 commands accordingly.

Todo
- [ ] support the new option in lint-command
- [ ] support the new option in tokenize-command

### DIFF
--- a/src/Utils/CLI.php
+++ b/src/Utils/CLI.php
@@ -31,7 +31,7 @@ class CLI
 
     public function usageHighlight()
     {
-        echo "Usage: highlight-query --query SQL [--format html|cli|text]\n";
+        echo "Usage: highlight-query [--query SQL|--path path/to/my.sql] [--format html|cli|text]\n";
     }
 
     public function getopt($opt, $long)
@@ -44,10 +44,11 @@ class CLI
         $longopts = [
             'help',
             'query:',
+            'path:',
             'format:',
         ];
         $params = $this->getopt(
-            'hq:f:',
+            'hq:p:f:',
             $longopts
         );
         if ($params === false) {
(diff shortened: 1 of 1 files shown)


src/Utils/CLI.php
```diff
@@ -31,7 +31,7 @@ public function mergeLongOpts(&$params, &$longopts)
 
     public function usageHighlight()
     {
-        echo "Usage: highlight-query --query SQL [--format html|cli|text]\n";
+        echo "Usage: highlight-query [--query SQL|--path path/to/my.sql] [--format html|cli|text]\n";
```
staabm: initially I named this option `--file`, but this somehow collides with the `--format` option when using `getopt` (I guess because of the same first character or something).

to get arround this naming problem I renamed the option to `--path`.

If you like `--file` more I can do the changes, after someone guided me on how the opts-format needs to look like.

staabm:
Just implemented the `stdin` reading case in https://github.com/phpmyadmin/sql-parser/pull/239

this might make this change here obsolete. please tell me your feeling whether supporting this parameter still makes sense, or whether people should just use
`cat my.sql | bin/highlight-query` or similar instead of `bin/highlight-query --path=my.sql`

devenbansod:
I'm in favor of the stdin approach in #239 (though I don't have a very strong opinion on it and could be convinced otherwise) because I find it sleeker and more pluggable over this approach.

staabm:
I totally agree.

devenbansod:
Thanks for opening  this PR.

I'm closing this and we'll continue the discussion at #239. 👍

ibennetch:
I think this is an awesome improvement and I’m really looking forward to getting it implemented. I don’t have a strong feeling about whether this or the other is a better approach, but Deven‘s reasoning seems sound. The stdin approach is more flexible for piping and I don’t see any compelling reason to favor this approach.

PR #241, opened 2019-04-10
Fix PHP warnings when building an incomplete CASE expression

LeSuisse:
Issue can be reproduced with the following code snippet:
```php
new \PhpMyAdmin\SqlParser\Parser('SELECT a CASE');
// PHP Warning:  count(): Parameter must be an array or an object that implements Countable in src/Components/CaseExpression.php on line 296
// PHP Warning:  count(): Parameter must be an array or an object that implements Countable in src/Components/CaseExpression.php on line 297
```


ibennetch:
I'm a little unsure about this because it doesn't seem to have fixed anything with the command line interface:

```
$ ./bin/highlight-query --query 'SELECT a CASE;'
SELECT
    a CASE;
$  ./bin/lint-query --query 'SELECT a CASE;'
#1: Unexpected end of CASE expression (near "" at position 0)
```

LeSuisse:
The PHP warnings have been fixed.

Without this contribution:
```
$ ./bin/lint-query --query 'SELECT a CASE;'
PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /home/thomas/Workspace/PHP/sql-parser/src/Components/CaseExpression.php on line 296
PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /home/thomas/Workspace/PHP/sql-parser/src/Components/CaseExpression.php on line 297
#1: Unexpected end of CASE expression (near "" at position 0)
```

With this contribution:
```
$ ./bin/lint-query --query 'SELECT a CASE;'
#1: Unexpected end of CASE expression (near "" at position 0)
```

This change is very similar to other fixes like the one done in 2b3fb2e5783e458f8ae07ccf9cd96f65df5cface

ibennetch:
Yes, this makes sense now. Good job.

I was initially testing with PHP 7.1 which does not give the warning, but with PHP 7.3 I see the same results you do.

Merging this, thanks for your contribution.

PR #243, opened 2019-04-28
Add usage hints for command line stdin

ibennetch:
Signed-off-by: Isaac Bennetch <bennetch@gmail.com>


devenbansod:
@ibennetch it seems we need to adjust the tests for CLI.php

ibennetch:
Thanks, that should be fixed now.

PR #245, opened 2019-05-07
Added null check

staabm:
a shot in the dark to fix:

> Error: Class name must be a valid object or a string

https://github.com/phpmyadmin/sql-parser/pull/224#issuecomment-489885276

### DIFF
--- a/src/Statement.php
+++ b/src/Statement.php
@@ -401,9 +401,11 @@ abstract class Statement
             $this->after($parser, $list, $token);
 
             // #223 Here may make a patch, if last is delimiter, back one
-            if ((new $class()) instanceof FunctionCall) {
-                if ($list->offsetGet($list->idx)->type === Token::TYPE_DELIMITER) {
-                    --$list->idx;
+            if ($class !== null) {
+                if ((new $class()) instanceof FunctionCall) {
+                    if ($list->offsetGet($list->idx)->type === Token::TYPE_DELIMITER) {
+                        --$list->idx;
+                    }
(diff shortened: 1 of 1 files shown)


src/Statement.php
```diff
@@ -401,9 +401,11 @@ public function parse(Parser $parser, TokensList $list)
             $this->after($parser, $list, $token);
 
             // #223 Here may make a patch, if last is delimiter, back one
-            if ((new $class()) instanceof FunctionCall) {
-                if ($list->offsetGet($list->idx)->type === Token::TYPE_DELIMITER) {
-                    --$list->idx;
+            if ($class !== null) {
+                if ((new $class()) instanceof FunctionCall) {
```
staabm: I guess this could be simplified using `is_a()` so we would not need to create a object just to check for its class

https://php.net/is_a

staabm:
//cc @ibennetch

ibennetch:
That sure looks like it fixes it! I'll let the whole Travis suite finish running first, but this works correctly on my local system. Nice work.

src/Statement.php
ibennetch: @staabm I think I'll merge this now to fix the current errors, and if you have time and a desire to modify the code to use `is_a()` instead, that would be a big benefit. Since we can implement the is_a() method in more places than just this one, I think it makes more sense to merge this and then make modifications. Sound good?

staabm: Please go ahead

PR #264, opened 2019-10-26
Manage "CHECK" keyword

Tithugues:
Even if this keyword is ignored by MySQL, it should be parsed.


williamdes:
Thank your for your contribution @Tithugues

PR #265, opened 2019-10-31
Fix #13951 wrong parsing partitions

Tithugues:
When partition names starts with a number, the parser wasn't reading the full name and stopped before the first "_".

### DIFF
--- a/src/Components/PartitionDefinition.php
+++ b/src/Components/PartitionDefinition.php
@@ -170,10 +170,16 @@ class PartitionDefinition extends Component
                 $ret->name = $token->value;
 
                 // Looking ahead for a 'VALUES' keyword.
-                $idx = $list->idx;
-                $list->getNext();
-                $nextToken = $list->getNext();
-                $list->idx = $idx;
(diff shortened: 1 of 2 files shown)


williamdes:
Can you please add a test case?

Btw we merge bug fixes on QA branch, so if you want to rebase :)

Tithugues:
Sorry, I created the PR on the wrong branch, but the commit were done on QA.

I'll see to update the unit tests.

ibennetch:
In that case, it’s quite easy to change the target branch here on GitHub (though I can’t do it right now as I’m on mobile).

Near the top of the page is an Edit button for changing the pull request title. If you cli k that, the branch name turns to a drop down selector.

Thanks for your contribution!

Tithugues:
@williamdes @ibennetch , test case is done.

Is this OK?

Tithugues:
I saw the failure in the unit test! I'll fix it and run the whole tests, not only the `PartitionDefinitionTest`!

Tithugues:
Should be better now...

PR #276, opened 2019-12-29
fix #255: add missing options in SET statement

### LINKED ISSUE #255: SET statement not properly parsed or build
```php
$sql    = 'SET SESSION   sql_mode = \'TRADITIONAL\'';
$parser = new Parser($sql);
$s2     = $parser->statements[0]->build();
// $s2 is **missing whitespace** in **SESSIONsql_mode**
//SET  SESSIONsql_mode = 'TRADITIONAL'

//PHP 7.1.18
//"name": "phpmyadmin/sql-parser",
//"version": "5.0.0",
```

devenbansod:
Ref: https://dev.mysql.com/doc/refman/8.0/en/set-variable.html

Not sure if this should target the QA branch or the master branch since it changes a static field (which could be called a breaking change by some).

@mauriciofauth @williamdes what do you think?


williamdes:
And QA branch is the right target, the difference is just php support and some code changes

Issue #270, opened 2019-11-24
sscanf() has been disabled for security reasons


williamdes:
https://github.com/phpmyadmin/sql-parser/blob/69ee1104925e7cdcc29269cd653ecb3ef6e71937/src/Token.php#L261-L271

Some people disable `sscanf` for some reasons

And this issue is appearing on our error reporting server since phpMyAdmin 4.5

PR #286, opened 2020-02-01
Replace sscanf by equivalent native PHP functions because sscanf can be disabled for security reasons.

niconoe-:
As sscanf can be disabled for security reasons, as mentioned in #270 , and also explained here: https://www.php.net/manual/en/function.sscanf.php#49713, I changed the usages by other native PHP functions, so performance should not be impacted.

Fixes: #270


src/Token.php
```diff
@@ -252,17 +252,16 @@ public function extract()
                 if ($this->flags & self::FLAG_NUMBER_HEX) {
                     if ($this->flags & self::FLAG_NUMBER_NEGATIVE) {
                         $ret = str_replace('-', '', $this->token);
-                        sscanf($ret, '%x', $ret);
-                        $ret = -$ret;
+                        $ret = -hexdec($ret);
```
williamdes:
```suggestion
                        $ret -= hexdec($ret);
```
Would this be more obvious?

williamdes:
Tests seem to fail https://travis-ci.org/phpmyadmin/sql-parser/jobs/644946170#L323

src/Token.php
niconoe-: I don't think it's the same behavior. If, for example, `$this->token` worth `-a0` (-16 in decimal), expected value for `$ret` is -16. If you do `$ret -= hexdec($ret)`, you're doing `$ret = a0 - 16`, so basically, `$ret = 0`.

williamdes: Yes, I did not read it right
Can you add a test so we are sure nobody removes the sign by mistake?

williamdes:
Can you run `./vendor/bin/phpcbf` and commit the coding standard fixes
The PR will be okay after that

williamdes:
Thank you @niconoe- that will reduce errors being reported to our server :tada:

PR #287, opened 2020-02-02
Fixing conflict in lexer when '*/*' was in a query.

### LINKED ISSUE #285: Parsing comments issues
Hi,
When I parse this valid SQL statement:
`SELECT */* comment */ FROM table_name` 
the parser divides `*/*` into two tokens: 

- 1st token with `*/` in value and type TYPE_COMMENT
- 2nd token with `*`  in value and type TYPE_OPERATOR

The expected behavior would be:
- 1st token with `*` in value and type TYPE_OPERATOR
- 2nd token with `/*`  in value and type TYPE_COMMENT

Do you agree on that, and if yes, would you like help on providing a PR?
Thanks a lot

niconoe-:
This is willing to fix #285


niconoe-:
> Perfect, could you add more tests?

Sure, but what kind of tests do you want me to provide? I already tested all valid use cases of "*/*" I know.

williamdes:
> > Perfect, could you add more tests?
> 
> Sure, but what kind of tests do you want me to provide? I already tested all valid use cases of `*/*` I know.

Maybe a new line, multiples wildcards 
What would be the behavior of a column named like a wildcard? backquoted vs not backquoted

I would like more tests on strange queries more that normal ones if possible

niconoe-:
Ok I'll provide that. Thank you for your feedback.

niconoe-:
I added more complex test cases in the provider of the lexer. Is this OK now?
Thanks a lot.

williamdes:
CI passed :heavy_check_mark:

williamdes:
Merged, thank you @niconoe- !

PR #289, opened 2020-02-03
Fix issue 288 - Lexer fails to detect "*" as a wildcard.

### LINKED ISSUE #288: Lexer fails to detect "*" as a wildcard.
Hi guys, thank you for this impressive tool!

I have noticed that the Lexer is currently not able to flag the "*" operator as a wildcard, and will only flag it as an arithmetic operator.

I would like to know if it is correct, in your opinion, to consider that "*" is a wildcard if it is followed by (ignoring whitespaces and comments) any of the following:
- `FROM` keyword, like in _SELECT * FROM_
- `USING` keyword like in _DELETE table_name.* USING_
- a comma `,` like in _SELECT *, field_name FROM or in DELETE foo.*, bar USING_
- a closing parenthesis `)`, like in _COUNT(*)_

Otherwise, we should consider "*" as an arithmetic operator. Would you agree on this?

If yes, I would be happy to provide a PR ;).

Actually, on a project of mine, I really need to distinguish the wildcard from the arithmetic operator, that's why I encountered this little issue.
Thank you!

niconoe-:
Fixes #288. 

Process is that after the Lexer does his work and created a list of tokens, and before returning back this list, it parses that list in order to change some flags for the `*` operator.

Doing this by the lexer because it's his job, but it would be too complicated and heavier in performance to directly detect the good flag for this operator. So the job of solving their ambiguity is done later, when the token list is created.

Lots of tests have been updated and 2 new providers came: first to check each case where operator `*` is arithmetic (all flags must be `1`), and second to check each case where operator `*` is a wildcard (all flags must be `16`). 
In those new providers, some tests are not performed beause of a bug related in #287 because of `*/*` string found. Once the PR fixing #287 merged into the good branches, one can just uncomment those tests.

If you feel to add more tests, please do so or tell me what do you want to test.

Finally, If that way of processing is OK for you guys, I think it can also solve the flag ambiguity of `=` ( `WHERE foo = bar` != `@i = 0`) using the same process.

Thanks a lot, and have a good review!

EDIT: Oh and, this time, @williamdes I did see to start from branch QA, so no cherry-pick to do ;).


williamdes:
It looks good
But our storing format is not readable on PRs, something needs to be done some day

I will review on my computer a second time

Thank you!

niconoe-:
> It looks good
> But our storing format is not readable on PRs, something needs to be done some day

I'm thinking of different ways to make them a little bit more readable, but maybe we should talk about it in another thread.

williamdes:
> I'm thinking of different ways to make them a little bit more readable, but maybe we should talk about it in another thread.

Could we discuss and change the format and then rebase your PRs so we can have a clear view on what changes ?
Or do you prefer that I merge your work but in that case we will not have another occasion to try in live the new diff

niconoe-:
I would prefer this PR being accepted because it is focused to what it's doing: to fix the lexer issue about the "*" as a wildcard. Review the format of the tests is something that should be part of a separated enhancement IMO.

williamdes:
Okay, I will do the merges soon
:fr:

williamdes:
Thank you @niconoe- 
Merged !

PR #290, opened 2020-02-06
Stop instanciating an object to check its class name.

### LINKED ISSUE #246: Use is_a() instead of creating an object  to check for its class
It's [been suggested](https://github.com/phpmyadmin/sql-parser/pull/245#pullrequestreview-234555279) by @staabm that we could use "`is_a()` so we would not need to create a object just to check for its class".

This seems like a good idea to me.

niconoe-:
Fixes #246 
When set on master, where PHP 5.3 is not maintained anymore, this could be a little bit improved.

I did not use `is_a` because of the function call (not good for performance) and I did not use `instanceof` neither because `$class` is a string, not an object. We are actually comparing 2 strings, so let's just compare 2 strings...

I'm aware it has not the exact same behavior in case of `$class` is a string evaluated to a class name of a child of `FunctionCall` class, but it appears that `FunctionCall` has no children, so the misbehavior here is not applicable.

### DIFF
--- a/src/Statement.php
+++ b/src/Statement.php
@@ -407,12 +407,11 @@ abstract class Statement
             $this->after($parser, $list, $token);
 
             // #223 Here may make a patch, if last is delimiter, back one
-            if ($class !== null) {
-                if ((new $class()) instanceof FunctionCall) {
-                    if ($list->offsetGet($list->idx)->type === Token::TYPE_DELIMITER) {
-                        --$list->idx;
-                    }
-                }
+            // TODO: when not supporting PHP 5.3 anymore, replace this by FunctionCall::class.
+            if ($class === 'PhpMyAdmin\\SqlParser\\Components\\FunctionCall'
+                && $list->offsetGet($list->idx)->type === Token::TYPE_DELIMITER
+            ) {
+                --$list->idx;
             }
         }
 


src/Statement.php
```diff
@@ -407,12 +407,11 @@ public function parse(Parser $parser, TokensList $list)
...
 
             // #223 Here may make a patch, if last is delimiter, back one
-            if ($class !== null) {
-                if ((new $class()) instanceof FunctionCall) {
-                    if ($list->offsetGet($list->idx)->type === Token::TYPE_DELIMITER) {
-                        --$list->idx;
-                    }
-                }
+            // TODO: when not supporting PHP 5.3 anymore, replace this by FunctionCall::class.
+            if ($class === 'PhpMyAdmin\\SqlParser\\Components\\FunctionCall'
```
williamdes: FunctionClass::class could not be more clean here?

williamdes: Oh, okay
TODO: when merging

niconoe-: Like I said in the `todo`, I just updated the QA branch which still supports PHP 5.3.

`FunctionCall::class` is a syntax only available since PHP 5.5.

williamdes: I keep the comment open so I do not forget to fix this on the QA>master merge

williamdes: f4d0d831d279005e46a76d7d8486c2071ec73e99

williamdes:
Thank you @niconoe- for this fix !

PR #300, opened 2020-04-13
Fix #16057 Don't allow numbers with letters

Tithugues:
Fix https://github.com/phpmyadmin/phpmyadmin/issues/16057

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -756,6 +756,7 @@ class Lexer extends Core
         //
         // Valid final states are: 2, 3, 4 and 6. Any parsing that finished in a
         // state other than these is invalid.
+        // Also, negative states are invalid states.
         $iBak = $this->last;
         $token = '';
         $flags = 0;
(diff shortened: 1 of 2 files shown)


Tithugues:
The `parseNumber` was stopping the parsing of a number when finding another character than ".", "e" or "E". But, if you have this kind of string: `0search`, this is not a number. That's what the `parseNumber` tries to detect now.

williamdes:
Okay, perfect then
Do you see any test case that could be added to cover the case where the parser should stop ?

Tithugues:
I'm sure that there are… But the test system with the `parse.out` files is so unusable that I should confess I'm a bit discouraged…

williamdes:
> I'm sure that there are… But the test system with the parse.out files is so unusable that I should confess I'm a bit discouraged…

I agree, It is impossible to read and view diffs
I hate that
I have a proposal to change the system would be using json_encode and json_decode

williamdes:
Merged !
Thank you !

PR #311, opened 2020-08-23
GH-310: Adds select validation to create view

### LINKED ISSUE #310: Linter accepts false syntax in view definitions
Thanks for this wonderful library. 

It seems like the linter doesn't catch any syntax errors for the view definitions.

Running `./vendor/bin/lint-query --query "CREATE VIEW abc AS WRONG_SELECT a, b FROM a"` gives back no parsing errors even though the syntax is clearly wrong.

Tested on the latest version `5.3`, the older `4.1` version also had the same issue.

aszenz:
The view statement body was not being  validated, used existing
select statement to ensure that the view definition is correctly
linted.
Also updated generated parser tests in / out tests.

Fixes #310


williamdes (approved):
LGTM :clap: :100:

src/Statements/CreateStatement.php
```diff
@@ -702,6 +702,7 @@ public function parse(Parser $parser, TokensList $list)
                 }
                 $this->body[] = $token;
             }
+            $this->select = new SelectStatement($parser, new TokensList($this->body));
```
devenbansod, line 705: @aszenz instead of creating a new `TokensList` here, could we not check if the first token is `SELECT` and then invoke `$this->select = new SelectStatement($parser, $list);` instead?

Right now, we would be using both `body` as well `select` which would be a duplication.

Moreover we should also fix the `build` function's path for CREATE VIEW statements to use this `$this->select` values if it's available. Please also add a test for the builder.

aszenz: Thanks for the review, initially I started on the same path , replacing the body with the select expression in both parse and build functions but some existing tests broke because of trimmed white space so I changed my approach to do limited changes only.
I'll take another look at it now, based on my initial understanding of the code.

aszenz:
>Left a comment/suggestion. 

Made some changes based on these

>Also, please rebase the branch over latest phpmyadmin/master.

This branch was originally made from QA, I switched both the pr and branch to master but now phpstan is complains about existing code and new code in the travis-ci. 
Should I fix the existing code as well in this branch?

williamdes:
> @aszenz I'm sorry for that earlier suggestion to rebase over master. I think QA might be the way to go.
> @williamdes could you help decide which branch to target for this?
> 
> If all the tests pass and coverage is good, I'm okay to merge it once we decide on the target branch.

Rebased (downbased) onto QA

williamdes:
Waiting on https://travis-ci.org/github/phpmyadmin/sql-parser/builds/721614968 now :)

src/Statements/CreateStatement.php
```diff
@@ -694,13 +694,26 @@ public function parse(Parser $parser, TokensList $list)
...
+            ) {
+                $list->idx = $nextidx;
+                $this->select = new SelectStatement($parser, $list);
+            } else {
+                for (; $list->idx < $list->count; ++$list->idx) {
+                    $token = $list->tokens[$list->idx];
+                    if ($token->type === Token::TYPE_DELIMITER) {
+                        break;
+                    }
+                    $this->body[] = $token;
```
williamdes, line 715: @aszenz Could you check this part of the code is still good after my rebase ?

aszenz: yeah still the same. thanks for rebasing

mauriciofauth (approved):

williamdes:
Thank you @aszenz for your contribution and patience !
This will be part of the next 4.7.0 and 5.4.0 releases

src/Statements/CreateStatement.php
```diff
@@ -694,13 +694,26 @@ public function parse(Parser $parser, TokensList $list)
                 $list->getNext();
             }
 
-            // Parsing the `AS` keyword.
-            for (; $list->idx < $list->count; ++$list->idx) {
-                $token = $list->tokens[$list->idx];
-                if ($token->type === Token::TYPE_DELIMITER) {
-                    break;
+            // Parsing the SELECT expression with and without the `AS` keyword
```
williamdes, line 697: Hi @aszenz 
Do you have an example of a create view without an `AS` ?

PR #312, opened 2020-08-27
Fix #309: Use stream_select instead of non-blocking STDIN

### LINKED ISSUE #309: testStdinPipe fails intermittently on s390x
In Ubuntu, on s390x (only), we've been noticing occasional failures with CLITest::testStdinPipe in phpmyadmin-sql-parser version 4.6.1.  This test runs the php binary six times.  Out of five different CI runs:

    A: data set #2 and #5 failed
    B: data set #2 and #3 failed
    C: data set #1 and #3 failed
    D: data set #3 and #4 failed
    E: data sets #0, #1, #2, #3, #4, #5 all failed

https://autopkgtest.ubuntu.com/packages/phpmyadmin-sql-parser/groovy/s390x

In the cases where the CI passed, we weren't doing anything special, it seems to have just randomly passed.

I unfortunately don't have access to the s390x machine so am unable to investigate this directly, just via the CI system.  The php cli and test suite pass properly on my (non-s390x) hardware.  I'm hoping you may have some insights to help us make the test more deterministic?

sergiodj:
The idiom used in the readStdin function from src/Utils/CLI.php is not
entirely error-prone: marking stdin as non-blocking and then reading
from it can lead to racy scenarios where we are not able to catch
what's being fed to the program.  This can be seen very frequently on
s390x, where a loop like this:

  $ while echo "invalid query" | /usr/bin/php7.4 /path/to//highlight-query; do sleep 1; done;

will lead to failures once on every few runs (as can be seen in the
description of issue #309).

This commit implements the more robust method which uses stream_select
to wait for stdin to become available, but resorts to a sensible
timeout value (0.2 seconds) which will prevent the blocking from
happening.

I tested this patch extensively on s390x and amd64, and noticed that
the non-determinism is gone.

Signed-off-by: Sergio Durigan Junior <sergio.durigan@canonical.com>

### DIFF
--- a/src/Utils/CLI.php
+++ b/src/Utils/CLI.php
@@ -235,9 +235,9 @@ class CLI
 
     public function readStdin()
     {
-        $read = [STDIN];
-        $write = [];
-        $except = [];
+        $read = array(STDIN);
+        $write = array();
+        $except = array();
 
         // Assume there's nothing to be read from STDIN.
         $stdin = null;
--- a/src/Utils/CLI.php
+++ b/src/Utils/CLI.php
@@ -233,11 +233,22 @@ class CLI
         return 1;
     }
 
-    public function readStdin() {
-        stream_set_blocking(STDIN, false);
-        $stdin = stream_get_contents(STDIN);
-        // restore-default block-mode setting
-        stream_set_blocking(STDIN, true);
+    public function readStdin()
+    {
+        $read = [STDIN];
+        $write = [];
+        $except = [];
+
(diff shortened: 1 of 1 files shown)


williamdes (approved):

williamdes:
Merged into QA as 0770895be14d06f3ee9df912e947d7774eb11872 and into master as 29ce64ec4421015fc0bd2f08a343786967cb5f5c

williamdes:
Thank you so much !

Issue #299, opened 2020-04-13
array_key_exists warning when parsing a "DEFAULT FALSE" token


devnix:
```
PHP Warning:  array_key_exists(): The first argument should be either a string or an integer in /var/www/html/vendor/phpmyadmin/sql-parser/src/Components/AlterOperation.php on line 286
```

A dump inside AlterOperation.php gives me the following result:
![image](https://user-images.githubusercontent.com/1777519/79115458-2c448880-7d86-11ea-900f-85dca0a8edcb.png)

I don't have any idea if the `value` attribute should be a boolean in this case, so I'm not starting a PR right now. I don't know if the correct solution would be to cast to string on the `array_key_exists` itself, or to save the `value` attribute as a string instead of a boolean...

onlime:
This still happens in latest PMA 5.0.3:

```
Warning in ./vendor/phpmyadmin/sql-parser/src/Components/AlterOperation.php#284
array_key_exists(): The first argument should be either a string or an integer
```

Any idea how this could be fixed?

williamdes:
I added this to a milestone to have a fix as quickly as possible.

shucon:
Starting to work on this issue.

shucon:
I'm not sure how to replicate the issue. Any steps to replicate would be appreciated. :smile:

onlime:
Here you go, @shucon, you can reproduce it with the following simple `.sql` file, stripped down from some of my customer's schema:

```sql
DROP TABLE IF EXISTS `searches`;
CREATE TABLE `searches` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(255) COLLATE utf8_unicode_ci NOT NULL,
  `public_name` varchar(120) COLLATE utf8_unicode_ci NOT NULL,
  `group_id` smallint(5) unsigned NOT NULL DEFAULT '0',
  `shortdesc` tinytext COLLATE utf8_unicode_ci,
  `show_separators` tinyint(1) NOT NULL DEFAULT '0',
  `deleted` tinyint(1) NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`)
) ENGINE=MyISAM  DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci ;

ALTER TABLE `searches` ADD `admins_only` BOOLEAN NOT NULL DEFAULT FALSE AFTER `show_separators`;
```

[...]

PR #314, opened 2020-10-12
Fixes array_key_exists warning when parsing a "DEFAULT FALSE" token

shucon:
Fixes: https://github.com/phpmyadmin/sql-parser/issues/299

The issue was caused because the value of token `FALSE` became boolean instead of string.

Signed-off-by: Saksham Gupta <shucon01@gmail.com>

### DIFF
--- a/src/Components/AlterOperation.php
+++ b/src/Components/AlterOperation.php
@@ -263,6 +263,12 @@ class AlterOperation extends Component
                 }
                 $state = 2;
             } elseif ($state === 2) {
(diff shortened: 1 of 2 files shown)


williamdes:
Could you add some phpunit tests to cover this cases ?

shucon:
Hey, I've added a test but I'm not sure whether that's how it's supposed to be.

ibennetch (approved):
I think this LGTM. Thank you.

PR #325, opened 2021-01-15
phpdoc use GroupKeyword for groups

soyuka:


williamdes:
Hi @soyuka 
I do not know how you did find out that but it looks valid and https://github.com/phpmyadmin/sql-parser/blob/4823c4fd928cc6ea41e0ca141d9c98eb11daf726/src/Parser.php#L189-L192 seems to confirm that

soyuka:
I'm using psalm and the following code was flagged as erroneous as `$statement->group` was typed as OrderKeyword:

```php
            $groups = [];
            foreach ($statement->order as $order) {
                $groups[] = new GroupKeyword($order->expr);
            }

            // Had to type this manually
            /** @var OrderKeyword[] */
            $statement->group = $groups;
```

Thanks for the quick merge!

williamdes:
Very cool, we use psalm on the main project. Let me know if you find other issues

Issue #331, opened 2017-08-25
Add support for WITH clause (enhancement for syntax highlighter)


stevenwdv:
Currently the SQL syntax highlighter marks the `WITH` keyword as `Unrecognized statement type`.
MariaDB supports a `WITH` clause in combination with a `SELECT` statement since version 10.2.1 (stable=10.2.8), see https://mariadb.com/kb/en/the-mariadb-library/with/

**phpMyAdmin version:** 4.7.4

Sort of relates to phpmyadmin/phpmyadmin#13615 (but not really)

williamdes:
Related to #165

sturtevant:
I believe this issue is also disrupting the way phpMyAdmin result pagination works. Any query using a CTE ("WITH" clause) will return all results to the browser (even while the interface says "Showing rows 0 - 24"). This typically results in an HTTP timeout for large tables / queries.

To reproduce (any table of sufficient length will do):
```
SELECT id FROM mytable; -- pagination works for this query

WITH my_cte AS ( SELECT id FROM mytable ) SELECT id FROM my_cte; -- pagination doesn't work (all rows returned)
```

soyuka:
Can we close this one as duplicate from https://github.com/phpmyadmin/sql-parser/issues/165 ?

williamdes:
> Can we close this one as duplicate from #165 ?

Now that a PR is done I would say this one can stay open

williamdes:
This will be part of 5.5.0
Implemented in: #334

Issue #165, opened 2017-08-26
WITH keyword


ibennetch:
The WITH keyword is now supported for CTEs starting with MariaDB 10.2.1 and MySQL 8.0.

See https://github.com/phpmyadmin/phpmyadmin/issues/13616

```
./lint-query --query "WITH t as (select a from t1 where b >= 'c') 
>   select * from t2,t where t2.c=t.a;"
#1: Unrecognized statement type. (near "WITH" at position 0)
```

timwis:
Hey, any updates on this by any chance? I'm still getting phpmyadmin warnings when using `WITH`

jucajuca:
same here... can't use WITH

pooryadb:
any changes ?!

ibennetch:
No, unfortunately this has not yet been added to phpMyAdmin (you can use the syntax, it just gives a warning message when you type it in).

brunocerq:
phpMyAdmin **4.9.5**
MariaDB **10.3.23**
Still have "**1 errors were found during analysis.
Unrecognized statement type. (near "WITH" at position 0)**"

When will this be implemented?

williamdes:
When someone will have given some of their free time to implement it ;)
Maybe you can have a try ?

brunocerq:
William, sorry if the question looked like a "question/demand", it was not.

williamdes:
This will be part of 5.5.0
Implemented in: #334

PR #334, opened 2021-03-11
Add WITH support

soyuka:
Fixes: #165
Fixes: #331

### DIFF
--- a/src/Components/WithKeyword.php
+++ b/src/Components/WithKeyword.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * `WITH` keyword builder.
+ */
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\SqlParser\Components;
+
+use PhpMyAdmin\SqlParser\Component;
+use PhpMyAdmin\SqlParser\Parser;
+use RuntimeException;
+
+/**
+ * `WITH` keyword builder.
+ */
+final class WithKeyword extends Component
+{
+    /** @var string */
+    public $name;
+
+    /** @var ArrayObj[] */
+    public $columns = [];
+
+    /** @var Parser */
+    public $statement;
+
+    public function __construct(string $name)
+    {
+        $this->name = $name;
+    }
+
+    /**
+     * @param WithKeyword $component
+     * @param mixed[]     $options
+     *
+     * @return string
+     */
+    public static function build($component, array $options = [])
+    {
+        if (! $component instanceof WithKeyword) {
+            throw new RuntimeException('Can not build a component that is not a WithKeyword');
+        }
+
+        if (! isset($component->statement)) {
+            throw new RuntimeException('No statement inside WITH');
+        }
+
+        $str = $component->name;
+
+        if ($component->columns) {
+            $str .= ArrayObj::build($component->columns);
+        }
+
+        $str .= ' AS (';
+
+        foreach ($component->statement->statements as $statement) {
+            $str .= $statement->build();
+        }
+
+        $str .= ')';
+
+        return $str;
+    }
+}
--- a/src/Statements/WithStatement.php
+++ b/src/Statements/WithStatement.php
@@ -1,0 +1,214 @@
+<?php
+/**
+ * `WITH` statement.
+ */
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\SqlParser\Statements;
+
+use PhpMyAdmin\SqlParser\Components\Array2d;
+use PhpMyAdmin\SqlParser\Components\OptionsArray;
+use PhpMyAdmin\SqlParser\Components\WithKeyword;
+use PhpMyAdmin\SqlParser\Exceptions\ParserException;
+use PhpMyAdmin\SqlParser\Parser;
+use PhpMyAdmin\SqlParser\Statement;
+use PhpMyAdmin\SqlParser\Token;
+use PhpMyAdmin\SqlParser\TokensList;
+use PhpMyAdmin\SqlParser\Translator;
+
+use function array_slice;
+use function count;
+
+/**
+ * `WITH` statement.
+
+ *  WITH [RECURSIVE] query_name [ (column_name [,...]) ] AS (SELECT ...) [, ...]
+ */
+final class WithStatement extends Statement
+{
+    /**
+     * Options for `WITH` statements and their slot ID.
+     *
+     * @var mixed[]
+     */
+    public static $OPTIONS = ['RECURSIVE' => 1];
+
+    /**
+     * The clauses of this statement, in order.
+     *
+     * @see Statement::$CLAUSES
+     *
+     * @var mixed[]
+     */
+    public static $CLAUSES = [
+        'WITH' => [
+            'WITH',
+            2,
+        ],
+        // Used for options.
+        '_OPTIONS' => [
+            '_OPTIONS',
+            1,
+        ],
+        'AS' => [
+            'AS',
+            2,
+        ],
+    ];
+
+    /** @var WithKeyword[] */
+    public $withers = [];
+
+    /**
+     * @param Parser     $parser the instance that requests parsing
+     * @param TokensList $list   the list of tokens to be parsed
+     */
+    public function parse(Parser $parser, TokensList $list)
+    {
+        ++$list->idx; // Skipping `WITH`.
+
+        // parse any options if provided
+        $this->options = OptionsArray::parse($parser, $list, static::$OPTIONS);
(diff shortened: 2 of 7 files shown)


soyuka:
About phpstan I'm unsure as I typed properties in a 7.4 fashion which threw on php 7.2 and now everything has phpdocs. Also, `@return void` doesn't seem to be used elsewhere. Let me know of I can fix these.

src/Statements/WithStatement.php
```diff
@@ -0,0 +1,204 @@
...
+        foreach ($this->withers as $wither) {
+            $str .= $str === 'WITH ' ? '' : ', ';
+            $str .= WithKeyword::build($wither);
+        }
+
+        return $str;
+    }
+
+    /**
+     * @return ParserException | TokensList
```
williamdes, line 164:
```suggestion
     * @return ParserException|TokensList
```

tests/Parser/WithStatementTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+use PhpMyAdmin\SqlParser\Components\WithKeyword;
+use PhpMyAdmin\SqlParser\Tests\TestCase;
+use PhpMyAdmin\SqlParser\Lexer;
+use PhpMyAdmin\SqlParser\Parser;
+use stdClass;
+
+class WithStatementTest extends TestCase
+{
+    public function testWith()
```
williamdes, line 15:
```suggestion
    public function testWith(): void
```

tests/Parser/WithStatementTest.php
```diff
@@ -0,0 +1,109 @@
...
+        // phpcs:disable Generic.Files.LineLength.TooLong
+        $expected = <<<SQL
+WITH categories(identifier, name, parent_id) AS (SELECT c.identifier, c.name, c.parent_id FROM category AS `c` WHERE c.identifier = 'a' UNION ALL SELECT c.identifier, c.name, c.parent_id FROM categories, category AS `c` WHERE c.identifier = categories.parent_id), foo AS (SELECT * FROM test)
+SQL;
+        // phpcs:enable
+        $this->assertEquals($expected, $parser->statements[0]->build());
+        $this->assertEquals('SELECT * FROM categories', $parser->statements[1]->build());
+    }
+
+    public function testWithHasErrors()
```
williamdes, line 44:
```suggestion
    public function testWithHasErrors(): void
```

williamdes: could you apply this to all methods of this class and update the phpstan baseline please ?

williamdes (approved):
Parfait !

tests/Parser/WithStatementTest.php
soyuka: It's not used in the other tests but sure I can add these and remove the phpstan baseline. Note that with latest phpstan it doesn't warn about these (php 8)

williamdes: Yup, this is more to have native type-hints on all tests. Thanks for updating them
If you can do some commit squashing that would be cool too :)

soyuka: you should be able to squash with the gh button but I'll do it

src/Statements/WithStatement.php
```diff
@@ -0,0 +1,204 @@
...
+
+use function array_slice;
+use function count;
+
+/**
+ * `WITH` statement.
+
+ *  WITH [RECURSIVE] query_name [ (column_name [,...]) ] AS (SELECT ...) [, ...]
+ */
+class WithStatement extends Statement
```
soyuka, line 28: should be final?

src/Statements/WithStatement.php
```diff
@@ -0,0 +1,204 @@
...
+                    continue;
+                }
+
+                if ($token->keyword === 'AS') {
+                    ++$list->idx;
+                    $state = 2;
+                    continue;
+                }
+            } elseif ($state === 2) {
+                if ($token->value === '(') {
```
soyuka, line 114: add a comment explaining what this does

src/Statements/WithStatement.php
```diff
@@ -0,0 +1,204 @@
...
+                        }
+                    }
+
+                    $this->withers[$wither]->statement = $subParser;
+                    continue;
+                }
+
+                if ($token->value === ',') {
+                    $list->idx++;
+                    $state = 0;
```
soyuka, line 136: comment

src/Statements/WithStatement.php
```diff
@@ -0,0 +1,204 @@
...
+                    continue;
+                }
+
+                // nothing else
```
soyuka, line 140: improve comment

src/Statements/WithStatement.php
williamdes: I would say no

Surfoo:
Hello,
When do you think it will be merged?

src/Components/WithKeyword.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
+ * `UNION` keyword builder.
```
dunglas, line 3:
```suggestion
 * `WITH` keyword builder.
```

src/Components/WithKeyword.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @param WithKeyword $component
+     * @param mixed[]     $options
+     *
+     * @return string
+     */
+    public static function build($component, array $options = [])
+    {
+        if (! $component instanceof WithKeyword) {
```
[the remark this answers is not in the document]
soyuka: Because I rely on the parent class typing I guess? (https://github.com/phpmyadmin/sql-parser/blob/a22dc4495c1e2e6f1caa3d1cb9586e21bcb1f77a/src/Component.php#L58)

src/Components/WithKeyword.php
```diff
@@ -0,0 +1,65 @@
...
+            throw new RuntimeException('Can not build a component that is not a WithKeyword');
```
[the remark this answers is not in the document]
soyuka: LogicException are more often used outside of running code, at compile time, here as we would hit that during the "runtime" it's more appropriate. See also https://stackoverflow.com/questions/5586404/logicexception-vs-runtimeexception. 

I hesitated here, and I even wanted to pick an exception of the package but ParseException wasn't a good fit.

src/TokensList.php
```diff
@@ -50,11 +50,7 @@ public function __construct(array $tokens = [], $count = -1)
         }
 
         $this->tokens = $tokens;
-        if ($count !== -1) {
-            return;
-        }
-
-        $this->count = count($tokens);
+        $this->count = $count === -1 ? count($tokens) : $count;
```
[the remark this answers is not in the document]
soyuka: Yes, there was a bug in the parser actually. `count` here should always be the value of the array's length.

soyuka:
20210512_10h01m39s_grim

`bin/lint-query` outputs nothing, error code 0

`bin/tokenize-query` works as well

williamdes:
Thank you :)
I will merge this pull-request this week-end or monday to get a chance to have a team member review it

devenbansod (approved):
Thanks for working on this @soyuka! LGTM 👍 💯

iifawzi:
Hi, while working on other issues related to the `WITH` statements, I walked through this implementation and I noticed that it's parsing the statement based on this definition https://dev.mysql.com/doc/refman/8.0/en/with.html
```
    WITH [RECURSIVE]
        cte_name [(col_name [, col_name] ...)] AS (subquery)
        [, cte_name [(col_name [, col_name] ...)] AS (subquery)] ...
```

I'm wondering, shouldn't we also parse the part that follows the `WITH` statements? which will always be either `SELECT` (mostly), `UPDATE` or `DELETE`, instead of depending on other statements to parse them? this would help in not duplicating the code as well as throwing an appropriate errors. I think we would only need to parse the tokens remained after the `WITH` tokens based on the statement type (select, update, or delete). And imo it makes sense because the `WITH` statements wouldn't work without them.

For now, the parser will not throw any error for such queries: 
```SQL
WITH cte (col1, col2) AS ( SELECT 1, 2 UNION ALL SELECT 3, 4 ) FROM cte;
```

what do you think? I might be missing something though.

williamdes:
It seems to make sense to handle this in WithStatement and not in another statement
Would it be difficult to implement?
The version will be released at the end of the week by the way

iifawzi:
Cool, I don't think it would be that difficult to implement, I will work on it. Hopefully I can manage to finish it along with the other issues related to it, before the end of the week.

PR #337, opened 2021-05-17
fix #168 - error on test SET statement

BorisValgo:
Fixes: #168

PR #343, opened 2021-08-14
Fixing the issue of ignoring the body tokens when creating views with `union`

iifawzi:
Hi, This PR should Fix https://github.com/phpmyadmin/phpmyadmin/issues/16549 and https://github.com/phpmyadmin/phpmyadmin/issues/16510

The issue that this commit https://github.com/phpmyadmin/sql-parser/commit/de566872af26fc8d785d110fcf3bc1890da73e8d introduced is that the code will never reach the `else` statement here: 
https://github.com/phpmyadmin/sql-parser/blob/184a80309d83e3cb6782696e4764bc15f60d9ec3/src/Statements/CreateStatement.php#L677-L693

because the `elseif` condition, will always be true - once we reach this entire `if` block embedded above, the token will be the `AS` keyword, which's always followed by `SELECT` - that's why the else statement will never be reached and all other tokens are ignored, that's being said, I've moved the `elseif` block to be part of the `else`, and modified the way of returning the `create view` statement to return both the main select query with the build of other body tokens if there're any. 

I've modified the tests and added another two. That's all I noticed while debugging, I might be missing something though. 

Signed-off-by: Fawzi E. Abdulfattah <iifawzie@gmail.com>


williamdes (approved):
This really looks good, thank you so much !
The diff on master branch in out files really makes sense

src/Statements/CreateStatement.php
```diff
@@ -700,14 +700,16 @@ public function parse(Parser $parser, TokensList $list)
                 && $token->keyword === 'SELECT'
             ) {
                 $this->select = new SelectStatement($parser, $list);
```
williamdes, line 702: this line is not covered for some reason, but I merged your work as it seems correct
Will try to integrate it locally with phpMyAdmin just to have an idea of if everything is still working okay

iifawzi: I couldn’t find a proper example for a view without the AS keyword, I don’t think that this’s even possible. 

https://mariadb.com/kb/en/create-view/

https://dev.mysql.com/doc/refman/8.0/en/create-view.html

If AS can be omitted and normal select queries could be made with joins, unions or whatever, I think we would need to make some changes because we will encounter the same issue where the body tokens will be ignored because the condition will always be true and the code will not reach the else statement. 

Anyway, I’ve integrated the changes with PMA, and everything worked as expected and all tests passed.

williamdes: Did some more research and found nothing but a new exotic syntax that seems to work fine: 08853017f783d73499bacc5bce3b53cb21511311

iifawzi: Awesome! so, are we sure that there wouldn't be any case that could start with anything other than `AS` keyword? 
if this's the case, maybe we can get rid of the main if statement

williamdes: I would say, remove it as it is dead code. If people can find a proof well we will add it back and make some proper code to handle it ;)

iifawzi: Great, do you wanna me to make the commit in this PR? or a separate one?

williamdes: I do not know how to thank you for all your awesome 🙏🏻  (do you have an online wishlist ?)
You can push a commit I will pick it

iifawzi: I've committed the change https://github.com/iifawzi/sql-parser/commit/24eed885811a62ea243a4bec9d460fe3928f3d29 and tests in both PMA, and SQL-Parser and all tests passed

It's my pleasure that I'm contributing to such a great project. Your help, encouraging replies, and the great discussions, is the reason behind all of this. Thank you too. 
unfortunately; I don't have any :d

williamdes: Yay !
Back to 100% coverage 🎉 
Maybe you would want to leave some thoughts on https://github.com/phpmyadmin/sql-parser/pull/337 or a patch ?

iifawzi: Sure, gonna take a look

Issue #168, opened 2017-09-04
SET statement / SET type


nijel:
It seems that the parser seems to treat SET as statement even when it should not:

```console
./bin/lint-query --query 'ALTER TABLE `table_name` ADD `new_column_name` SET ('value1','value2','value3') NOT NULL AFTER `another_column`'
#1: A new statement was found, but no delimiter between it and the previous one. (near "SET" at position 47)
#2: Unrecognized alter operation. (near "," at position 65)
#3: Unrecognized alter operation. (near "" at position 0)
```

See https://github.com/phpmyadmin/phpmyadmin/issues/13636

kapersoft:
I am also having troubles with the SET-expression in ALTER TABLE-queries:

```php
$query = "ALTER TABLE `table_name` ADD `new_column_name` SET ('value1','value2','value3') NOT NULL AFTER `another_column`";
$parser = new \PhpMyAdmin\SqlParser\Parser($query);
var_dump($parser->statements);
```

In the var-dump the `SET'`-, `(`-expressions and the first value between parentheses `'value1'` are missing.

I am willing to dig into this problem and submit a PR. Any ideas where to start? 

```php
array(1) {
  [0] =>
  class PhpMyAdmin\SqlParser\Statements\AlterStatement#31 (5) {
    public $table =>
    class PhpMyAdmin\SqlParser\Components\Expression#33 (7) {
      public $database =>
      NULL
      public $table =>
      string(10) "table_name"
      public $column =>
      NULL
      public $expr =>
      string(12) "`table_name`"
      public $alias =>
      NULL
      public $function =>
      NULL
      public $subquery =>
      NULL
    }
    public $altered =>
    array(3) {
      [0] =>
      class PhpMyAdmin\SqlParser\Components\AlterOperation#34 (3) {
        public $options =>
        class PhpMyAdmin\SqlParser\Components\OptionsArray#35 (1) {
          public $options =>
          array(1) {
            [1] =>
            string(3) "ADD"
          }
        }
        public $field =>
        class PhpMyAdmin\SqlParser\Components\Expression#36 (7) {
          public $database =>
          NULL
          public $table =>
          NULL
          public $column =>
          string(15) "new_column_name"
          public $expr =>
          string(17) "`new_column_name`"
          public $alias =>
          NULL
          public $function =>
          NULL
          public $subquery =>
          NULL
        }
        public $unknown =>
        array(0) {
        }
      }
      [1] =>
[...]

PR #344, opened 2021-08-15
Fixing parser errors when adding columns with SET type

iifawzi:
Hi, This PR should fix https://github.com/phpmyadmin/sql-parser/issues/168 and the issues shown https://github.com/phpmyadmin/sql-parser/pull/337

I will clarify the changes in a review.

Testing in PMA too.


src/Components/AlterOperation.php
```diff
@@ -282,13 +282,42 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
+
+                        if (
+                            ($token->value !== 'SET' ||
+                            (
+                                $list->tokens[$currentTokenID - 1]->value !== 'CHARACTER' &&
+                                $nextToken->value !== '(' &&
+                                $nextToken->value !== 'DEFAULT'
+                            )
+                            )
+                        ) {
```
iifawzi, line 303: This part is added to fix the issue of determining the next keyword without assuming the number of spaces as shown @168, any number of spaces between the `SET` keyword and either `(` or `DEFAULT` shouldn't cause any errors now.

src/Components/AlterOperation.php
```diff
@@ -282,13 +282,42 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
+                        if ($nextToken->value === '(') {
+                            // To avoid adding the tokens between the SET() parentheses to the unknown tokens
+                            $list->getNextOfTypeAndValue(Token::TYPE_OPERATOR, ')');
+                        } elseif ($nextToken->value === 'DEFAULT') {
+                            // to avoid adding the `DEFAULT` token to the unknown tokens.
+                            ++$list->idx;
+                        } else {
+                            // Reverting the changes we made in the beginning
+                            $list->idx = $currentTokenID;
+                        }
```
iifawzi, line 320: This part is added to fix the issue of adding many `unknown` tokens. The values between the parentheses in the case of using `SET()` will not be considered as unknown tokens,  and the `DEFAULT` keyword will not be considered unknown too. 

Now, things should work as usual, all other tokens that are following the column definition ( NOT NULL, AFTER, .... ) and the column type will be considered as `unknown`, and this's the case in all other tests. 

The added unknown tokens in the modified tests that include `SET` appeared because previously the error  was preventing the unknown tokens from being added as usual.

iifawzi (commented):
That's what I've noticed, I've also added some comments in the code, to make things clearer in the future

src/Components/AlterOperation.php
```diff
@@ -282,13 +282,42 @@ public static function parse(Parser $parser, TokensList $list, array $options =
                         // We have reached the end of ALTER operation and suddenly found
                         // a start to new statement, but have not find a delimiter between them
 
-                        if (! ($token->value === 'SET' && $list->tokens[$list->idx - 1]->value === 'CHARACTER')) {
+                        // We want to get the next non-comment and non-space token after $token
+                        // therefore, the first getNext call will start with the current $idx which's $token,
+                        // will return it and increase $idx by 1, which's not gurantted to be non-comment
```
williamdes, line 287:
```suggestion
                        // will return it and increase $idx by 1, which's not guaranteed to be non-comment
```

iifawzi: I will commit and squash the commits, after finishing the reviews

src/Components/AlterOperation.php
```diff
@@ -282,13 +282,42 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
+                        // In order to get back to the original value of idx, we kept it in $currentID
+
+                        $currentTokenID = $list->idx;
+                        $list->getNext();
+                        $nextToken = $list->getNext();
+
+                        if (
+                            ($token->value !== 'SET' ||
+                            (
+                                $list->tokens[$currentTokenID - 1]->value !== 'CHARACTER' &&
```
williamdes, line 298: If I understand well this is the previous token `$list->tokens[$currentTokenID - 1]`

And here it is checked that we did not find a SET or that we did not find CHARACTER (does this work for COLLATE ?) followed with something that was not ( and not DEFAULT

I am trying to understand this block as it is the harder to figure out in this PR ^^

src/Components/AlterOperation.php
williamdes: > I will commit and squash the commits, after finishing the reviews

😍

src/Components/AlterOperation.php
```diff
@@ -282,13 +282,42 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
                         }
+
+                        if ($nextToken->value === '(') {
+                            // To avoid adding the tokens between the SET() parentheses to the unknown tokens
+                            $list->getNextOfTypeAndValue(Token::TYPE_OPERATOR, ')');
+                        } elseif ($nextToken->value === 'DEFAULT') {
+                            // to avoid adding the `DEFAULT` token to the unknown tokens.
+                            ++$list->idx;
+                        } else {
+                            // Reverting the changes we made in the beginning
```
williamdes, line 318: maybe could the throws be moved here as it could make it less specific as the if and elseif would guard two cases ?

Edit: or maybe split the exception in two and create more specific messages ?

What do you think about that ?

src/Components/AlterOperation.php
iifawzi: > If I understand well this is the previous token $list->tokens[$currentTokenID - 1]
Yes, it's pointing to the previous token, this specific part of checking `CHARACTER` was already written and I've struggled too, to understand what's intended from it, as per my understand it checks whether the `SET` keyword is preceded by `CHARACTER` or not. for CHARACTER SET modifications  
So if the SET is not followed by ( either `DEFAULT` or `(`  )  OR preceded by (`CHARACTER`) it will throw `A new statement.......`

src/Components/AlterOperation.php
iifawzi: I think we will still need to guard this two cases, because each behaves differently, If I got your points correctly. Anyway I was written it in a waaaay different way, and phpcbf decided to clean things up :d 

could you elaborate more on the idea of splitting the. exceptions?

src/Components/AlterOperation.php
williamdes: > for CHARACTER SET modifications

😱 I did not even think about it, oh crap that should be written more explicitly

> So if the SET is not followed by ( either DEFAULT or ( ) OR preceded by (CHARACTER) it will throw A new statement.......

Oh, I now understand
Could you please split that in two ifs so nobody struggles anymore reading it 😄 ?

src/Components/AlterOperation.php
williamdes: > could you elaborate more on the idea of splitting the. exceptions?

I posted a message above, but this was all about making one clear case for "CHARACTER SET modifications" and the other one if it was not a "CHARACTER SET modification"

src/Components/AlterOperation.php
iifawzi: Tbh, I struggled understanding it until you asked, created some logs to know which test will make this condition true, to figure out :D.
I will split them and squash the commits

iifawzi: Should we create a specific message for each if? if yes, what error message be appropriate then?  @williamdes, in both cases if not followed by default / ( or preceded by char... the statement will be invalid

williamdes: If possible to keep the same message then go for it as it will be less work, else keep it the same and when you push I will make suggestions based on the code reading :)

iifawzi: Committed the changes, couldn't think of a cleaner way

williamdes (approved):
💯 Oh yeah, this looks neat and readable
Codecov seems to complain about a non covered line for some reason

williamdes:
While trying to create some particular query I found this one made by phpMyAdmin on table charset change
```sql
ALTER TABLE `d` DEFAULT CHARSET=hp8 COLLATE hp8_english_ci;
```
Errors:
```
#1: Missing comma before start of a new alter operation. (near "DEFAULT CHARSET" at position 16)
#2: Unrecognized alter operation. (near "DEFAULT CHARSET" at position 16)
```

Keeping this to add a new test before merging

And another bunch of them

```
$ ./bin/lint-query --query 'ALTER TABLE `d` CHARSET=hp8'
#1: Unrecognized alter operation. (near "" at position 0)
$ ./bin/lint-query --query 'ALTER TABLE `d` CHARSET=hp8;'
#1: Unrecognized alter operation. (near ";" at position 27)
$ ./bin/lint-query --query 'ALTER TABLE tbl_name CONVERT TO CHARACTER SET charset_name;'
$ ./bin/lint-query --query 'ALTER TABLE `d` CHARACTER SET utf8;'
#1: Unrecognized alter operation. (near ";" at position 34)
$ ./bin/lint-query --query 'ALTER TABLE `d` CHARACTER SET utf8 COLLATE utf8_general_ci;'
#1: Unrecognized alter operation. (near ";" at position 58)
```

Would you mind if I add them into QA and then you rebase/squash ontop of it ?

iifawzi:
Interesting, gonna figure out how. 

> Would you mind if I add them into QA and then you rebase/squash ontop of it ?
Sure, feel free to add them.

williamdes:
> Interesting, gonna figure out how.
> 
> > Would you mind if I add them into QA and then you rebase/squash ontop of it ?
> 
> Sure, feel free to add them.

Done as 09fefbfa78385f53f1bac8376ceb77f7c04196db

iifawzi:
It seems like all new added tests aren't related to the changes we introduced in this PR. 
They are all caused because of
https://github.com/phpmyadmin/sql-parser/blob/b18edd6d1e6766bbc806f29bd1a553a7ef9ac753/src/Components/AlterOperation.php#L311-L313

looks like the fix should be somewhere here; because it return empty array. I will try to investigate more :')
https://github.com/phpmyadmin/sql-parser/blob/b18edd6d1e6766bbc806f29bd1a553a7ef9ac753/src/Components/AlterOperation.php#L236

Also, if i'm not wrong, `parseAlterTableCharacterSet4.in` is passing with no errors.

src/Components/AlterOperation.php
```diff
@@ -279,16 +279,35 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
+                        // will return it and increase $idx by 1, which's not guaranteed to be non-comment
+                        // and non-space, that's why we're calling getNext again.
+                        // In order to get back to the original value of idx, we kept it in $currentID
 
-                        if (! ($token->value === 'SET' && $list->tokens[$list->idx - 1]->value === 'CHARACTER')) {
+                        $currentTokenID = $list->idx;
+                        $list->getNext();
+                        $nextToken = $list->getNext();
+
+                        if ($token->value === 'CHARACTER SET'){
```
iifawzi, line 292: While investigating about the added tests, I've noticed that `CHARACTER SET` is always a single token. I'm not really sure though :')

williamdes: This is quite interesting
would that mean that the original code was not working correctly 🤔

After some sleep hours I will be back to your work :)

iifawzi: Good morning

> would that mean that the original code was not working correctly 

it seems like yes; for example the tokens that will be generated for `ALTER TABLE  'd' CHARACTER SET utf8 COLLATE utf8_general_ci;`
are: 

```JSON
{
   "tokens":[
      {
         "token":"ALTER",
         "value":"ALTER",
         "keyword":"ALTER",
         "type":1,
         "flags":3,
         "position":0
      },
      {
         "token":" ",
         "value":" ",
         "keyword":null,
         "type":3,
         "flags":0,
         "position":5
      },
      {
         "token":"TABLE",
         "value":"TABLE",
         "keyword":"TABLE",
         "type":1,
         "flags":3,
         "position":6
      },
      {
         "token":" ",
         "value":" ",
         "keyword":null,
         "type":3,
         "flags":0,
         "position":11
      },
      {
         "token":"`d`",
         "value":"d",
         "keyword":null,
         "type":8,
         "flags":2,
         "position":12
      },
      {
         "token":" ",
         "value":" ",
         "keyword":null,
         "type":3,
         "flags":0,
         "position":15
      },
      {
         "token":"CHARACTER SET",
         "value":"CHARACTER SET",
         "keyword":"CHARACTER SET",
         "type":1,
         "flags":7,
         "position":16
      },
      {
         "token":" ",
         "value":" ",
         "keyword":null,
         "type":3,
         "flags":0,
         "position":29
      },
      {
         "token":"utf8",
         "value":"utf8",
         "keyword":null,
         "type":0,
         "flags":0,
         "position":30
      },
      {
         "token":" ",
         "value":" ",
         "keyword":null,
         "type":3,
         "flags":0,
         "position":34
      },
      {
         "token":"COLLATE",
         "value":"COLLATE",
         "keyword":"COLLATE",
         "type":1,
         "flags":3,
         "position":35
      },
      {
         "token":" ",
         "value":" ",
         "keyword":null,
         "type":3,
         "flags":0,
         "position":42
      },
      {
         "token":"utf8_general_ci",
         "value":"utf8_general_ci",
         "keyword":null,
         "type":0,
         "flags":0,
         "position":43
      },
      {
         "token":";",
         "value":";",
         "keyword":null,
         "type":9,
         "flags":0,
         "position":58
      },
      {
         "token":null,
         "value":null,
         "keyword":null,
         "type":9,
         "flags":0,
         "position":null
      }
   ]
}

```

williamdes: Hello !

Cool, happy that it fixes even more bugs
I will be merging all this if I do not detect any more issues

iifawzi: All issues should be fixed now, could you take a look @williamdes

src/Components/AlterOperation.php
```diff
@@ -122,7 +124,9 @@ class AlterOperation extends Component
         'PRIMARY KEY' => 2,
         'SPATIAL' => 2,
         'TABLESPACE' => 2,
-        'INDEX' => 2
+        'INDEX' => 2,
+
+        'CHARACTER SET' => 3,
```
williamdes, line 129: No sure why you did add a space :)

src/Components/AlterOperation.php
```diff
@@ -279,16 +283,35 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
+                        // In order to get back to the original value of idx, we kept it in $currentID
+
+                        $currentTokenID = $list->idx;
+                        $list->getNext();
+                        $nextToken = $list->getNext();
 
-                        if (! ($token->value === 'SET' && $list->tokens[$list->idx - 1]->value === 'CHARACTER')) {
+                        if ($token->value === 'CHARACTER SET'){
+                        // Reverting the changes we made in the beginning
+                        $list->idx = $currentTokenID;
```
williamdes, line 298:
```suggestion
                            // Reverting the changes we made in the beginning
                            $list->idx = $currentTokenID;
```
Some spacing is missing :)

src/Components/AlterOperation.php
iifawzi: Just For better readability, there’s a space between 1 and 2 too. Can be removed

src/Components/AlterOperation.php
iifawzi: Gonna commit and squash after finishing your review

src/Components/AlterOperation.php
williamdes: Oh okay, nevermind then ;)

src/Components/AlterOperation.php
williamdes: Thanks, back to working on this PR, will test this in minutes and do fake merges to see how it looks

iifawzi: Awesome, I've squashed the last two commits, kept the first commit not squashed for better commits history, since they're kinda fixing different issues, if you want me to squash all the commits, let me know.

williamdes: All looks good ! 👏🏻

williamdes (approved):
🚢

src/Components/AlterOperation.php
iifawzi: I think this line is not even needed, character set queries will work because of the last commit, and that’s why the coverage complain about it, I think it can be removed safely. @williamdes

iifawzi: I’m not home, I will try removing it and run the tests to make sure that no tests will fail. But since the coverage is complaining about it, and we do have many tests including character set, so it means it’s not being used.

williamdes:
phpstan complains about this possible null value, will add a check
```
 ------ ------------------------------------------------------------------- 
  Line   src/Components/AlterOperation.php                                  
 ------ ------------------------------------------------------------------- 
  302    Cannot access property $value on PhpMyAdmin\SqlParser\Token|null.  
  305    Cannot access property $value on PhpMyAdmin\SqlParser\Token|null.  
 ------ ------------------------------------------------------------------- 
```

williamdes:
All tools pass but phpcs is nuts: https://github.com/phpmyadmin/sql-parser/runs/3344576860#step:7:10

williamdes:
All good, but unable to have the last line covered `if ($token->value === 'CHARACTER SET') {` searching

src/Components/AlterOperation.php
williamdes: Indeed !
I removed the is case in fd1552e54c43203affed06f07290c7022d39331a and everything continued to pass 🎉

iifawzi:
Happy to see everything fixed! 🥳🥳❤️

Issue #339, opened 2021-07-27
Bad procedure body


williamdes:
```sql
 CREATE DEFINER=`root`@`%` PROCEDURE `test2`(IN `_var` INT) NOT DETERMINISTIC NO SQL SQL SECURITY INVOKER NO SQL SQL SECURITY INVOKER SELECT _var
 ```
 
The detected is 
```sql
NO SQL
    SQL SECURITY INVOKER
SELECT _var
```
It should be
```sql
SELECT _var
```

Ref: https://github.com/phpmyadmin/phpmyadmin/issues/15991

williamdes:
https://dev.mysql.com/doc/refman/8.0/en/create-procedure.html
https://mariadb.com/kb/en/create-function/
https://mariadb.com/kb/en/create-procedure/

This seems like non sense: https://github.com/phpmyadmin/sql-parser/blob/6186c3fe51d887c0d815b752ddf0ffc6fb0ba264/src/Statements/CreateStatement.php#L50

and real full definitions:
https://github.com/phpmyadmin/sql-parser/blob/6186c3fe51d887c0d815b752ddf0ffc6fb0ba264/src/Statements/CreateStatement.php#L232

If you can @iifawzi , could you also confirm this looks not right to have it in $OPTIONS because it will not be found in the same place as DEFINER

williamdes:
Never mind, this is for `CREATE VIEW`

iifawzi:
> If you can @iifawzi , could you also confirm this looks not right to have it in $OPTIONS because it will not be found in the same place as DEFINER

I think this's incorrect yes, because the definer will never be in the same token

williamdes:
🤔 I am not sure to understand why FUNC_OPTIONS exists and seems to be not working at all

What makes `SQL SECURITY` glued and some other not glued together ?

<details><summary>token list from tests</summary>
<p>

```
RPhpMyAdmin\SqlParser\TokensList Object
(
    [tokens] => Array
        (
            [0] => PhpMyAdmin\SqlParser\Token Object
                (
                    [token] => CREATE
                    [value] => CREATE
                    [keyword] => CREATE
                    [type] => 1
                    [flags] => 3
                    [position] => 0
                )

            [1] => PhpMyAdmin\SqlParser\Token Object
                (
                    [token] =>  
                    [value] =>  
                    [keyword] => 
                    [type] => 3
                    [flags] => 0
                    [position] => 6
                )

[...]

PR #345, opened 2021-08-17
Fix #342 - Adding the support of altering user options

### LINKED ISSUE #342: Valid ALTER USER statement throws validation error "unrecognized alter operation"
### Describe the bug

When writing an `ALTER USER` statement in phpMyAdmin, a validation error is thrown, even though the query executes successfully.

### To Reproduce

Put the following query into the phpMyAdmin SQL query editor.

```
ALTER USER trevor REQUIRE NONE;
```

### Expected behavior

Query should not show any validation errors in the phpMyAdmin SQL editor.

### Screenshots

image
image

### Server configuration

- Operating system: Windows 10 w/ Docker Desktop
- Web server: Apache
- Database version: MySQL 8.0.26
- PHP version: 
- phpMyAdmin version: 5.1.1

### Client configuration

- Browser: Firefox 91
- Operating system: Windows 10

iifawzi:
Hi, this PR should add the support of altering user options, not all the options are included though, because I'm not aware of all of them, I've just introduced the base. 

Fixes #342
Signed-off-by: Fawzi E. Abdulfattah <iifawzie@gmail.com>


williamdes (approved):
This looks good, to better understand I will cut out and merge into QA and master your newly added tests

williamdes (approved):
Awesome !
I really like the diff it makes on master branch out files

src/Components/AlterOperation.php
```diff
@@ -129,6 +129,42 @@ class AlterOperation extends Component
         'CHARACTER SET' => 3,
     );
 
+    /**
+     * All user options.
+     *
+     * @var array
+     */
+    public static $USER_OPTIONS = array(
+        'ATTRIBUTE' => array(
```
williamdes, line 138: MySQL 8.0 also has `COMMENT` https://dev.mysql.com/doc/refman/8.0/en/alter-user.html

williamdes: Let me know if you want to add it or should I merge ;)

iifawzi: Sorry for being late, I will add it in minutes and add a test for it too.

williamdes: Oh no worries, I am working on another part of the parser

src/Components/AlterOperation.php
```diff
@@ -129,6 +129,46 @@ class AlterOperation extends Component
...
+            1,
+            'var'
+        ),
+        'COMMENT' => array(
```
williamdes, line 142: How do you determine the numbers and the format ?

I am working on https://github.com/phpmyadmin/sql-parser/issues/339 and https://github.com/phpmyadmin/sql-parser/blob/24eed885811a62ea243a4bec9d460fe3928f3d29/src/Statements/CreateStatement.php#L675 seems to not work at all

williamdes (approved):
🚢

src/Components/AlterOperation.php
iifawzi: Done, added, and squashed

src/Components/AlterOperation.php
iifawzi: For example; 
``` CHANGE FAWZI NAME```

there's 3 keywords in this query, so each one MUST have a different number.. that's it

iifawzi: I think here we want to skip these parameters, with maybe something like getNextOfType(`typehere`, 'SELECT'), not sure though that this's the issue, I will take a look and share my thoughts here

iifawzi: going to take a look and share my thoughts here

PR #349, opened 2021-08-21
Fix #348 - Add missing KEY options for MySQL and MariaDB

### LINKED ISSUE #348: Missing KEY options
Ref: https://dev.mysql.com/doc/refman/8.0/en/create-table.html
Ref: https://mariadb.com/kb/en/create-table/

williamdes:
Fixes: #348

### DIFF
--- a/src/Components/Key.php
+++ b/src/Components/Key.php
@@ -31,7 +31,7 @@ class Key extends Component
     public static $KEY_OPTIONS = array(
         'KEY_BLOCK_SIZE' => array(
             1,
-            'var',
+            'var=',
         ),
         'USING' => array(
             2,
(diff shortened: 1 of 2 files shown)


src/Components/Key.php
```diff
@@ -44,7 +44,26 @@ class Key extends Component
...
-        )
+        ),
+        'INVISIBLE' => 1,
+        'VISIBLE' => 1,
+        'ENGINE_ATTRIBUTE' => array(
+            3,
+            'var=',
+        ),
+        'SECONDARY_ENGINE_ATTRIBUTE' => array(
+            4,
```
williamdes, line 55: could this be `1` or anything ?
Are the numbers unique by token ?

iifawzi: It could be anything with taking care of using different numbers for all the tokens that shouldn’t be present with this token ( contradictory tokens must use different numbers ) 

Also, I’ve noticed once that the numbers control what tokens will be present first in the query after the build. So make sure that the output is identical to the input, then everything should be ok. 

I will try to dig more into this later this week to maybe know how exactly it works and the differences, from the code.

williamdes: Hi @iifawzi 
Did you have some chance to dig in the code ?

williamdes: Hi @iifawzi 
Did you have some chance to dig in the code ?

iifawzi: Hi, Sorry for the late reply @williamdes. Yes I do, I will dig more on it, in the end of the week.

PR #352, opened 2021-09-06
Add phpbench

williamdes:
- Add phpbench
- improve a function that costs a lot on phpMyAdmin imports (pushed onto master as 79d5db8474a029539424f2e643b90a3858862b37)

## Todo

- [x] `composer phpbench/phpbench requires php 7.2`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
         continue-on-error: ${{ matrix.experimental }}
         strategy:
             matrix:
-                php-version: ["7.1", "7.2", "7.3", "7.4", "8.0", "8.1", "8.2"]
+                php-version: ["7.2", "7.3", "7.4", "8.0", "8.1", "8.2"]
                 os: [ubuntu-latest]
                 experimental: [false]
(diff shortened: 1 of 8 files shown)


.github/workflows/tests.yml
```diff
@@ -56,3 +56,42 @@ jobs:
...
               with:
                   cli-args: "--format=php-clover build/logs/clover.xml"
+
+    benchmark-php:
+        name: Benchmark on php ${{ matrix.php-version }} and ${{ matrix.os }}
+        runs-on: ${{ matrix.os }}
+        continue-on-error: ${{ matrix.experimental }}
+        strategy:
+            matrix:
+                php-version: ["7.1", "7.2", "7.3", "7.4", "8.0"]
```
mauriciofauth, line 66: Is it necessary to run the benchmark on every PHP version?

williamdes: Not really, but this would be better to assure the same quality level on each php version

As some versions can have very different behaviors for the same function

williamdes:
image

I think we can drop PHP 7.1 @MauricioFauth ?

PR #369, opened 2022-01-09
Add `SqlModes` class with SQL modes constants

mauriciofauth:
- Creates the `PhpMyAdmin\SqlParser\SqlModes` class with the `Context::SQL_MODE_*` constants.
- Deprecates the `Context::SQL_MODE_*` constants.

I'm not sure if this is a good thing since an SQL mode can be different depending on the context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add missing return types annotations
 * Improve the WITH statements parser (#363)
+* Add SqlModes class with SQL modes constants
 
 ## [5.5.0] - 2021-12-08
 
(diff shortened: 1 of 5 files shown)


williamdes (approved):
> I'm not sure if this is a good thing since an SQL mode can be different depending on the context.

I am not sure to understand

mauriciofauth:
> > I'm not sure if this is a good thing since an SQL mode can be different depending on the context.
> 
> I am not sure to understand

Some MySQL/MariaDB's SQL modes differ depending on the version. This differences are not handled by sql-parser. I'm not sure if this is relevant or not to sql-parser.

williamdes:
> > > I'm not sure if this is a good thing since an SQL mode can be different depending on the context.
> > 
> > 
> > I am not sure to understand
> 
> Some MySQL/MariaDB's SQL modes differ depending on the version. This differences are not handled by sql-parser. I'm not sure if this is relevant or not to sql-parser.

Maybe relevant, but for now we just store and handle the declared modes so nothing more to do I think

PR #379, opened 2022-07-04
Update .gitattributes

voku:
Do not copy the test directory into the project, if we install it via composer.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+/tests export-ignore
 /tools export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore


williamdes:
I can not approve this because Debian uses the `tests/` folder to run autopkg tests: https://salsa.debian.org/phpmyadmin-team/sql-parser/-/blob/debian/latest/debian/tests/phpunit
Output: https://salsa.debian.org/phpmyadmin-team/sql-parser/-/jobs/2353133

williamdes:
> @williamdes Understand. Do you think we could use the "exclude"-setting in the composer.json here? thinking

Does it really work ?
I accept to add this, can you edit your PR ?

Yurunsoft:
OK.
Also I would like to ask when you can release a new version, it's been almost a year since the last release.
I've encountered a bug and I'm currently using the `dev-master` version.

williamdes:
This was released right now 🎉

PR #384, opened 2022-08-08
Change error handling from throwing an exception to returning a boolean value

annexus254:
In case an error occurs during loading of a context, let the Context::load()
method return a boolean value indicating whether the context was successfully
loaded or not.
Throwing an Exception causes a false positive in debuggers,since the Context::load()
method does the correct thing anyway


williamdes (approved):
Thank you for your contribution !
I rebased it onto `6.0.x` that will be the next major version to be released, since this change may be a breaking change for some users.
I also fixed the test for this and added a CHANGELOG.md  entry

MauricioFauth (approved):

Issue #378, opened 2022-06-13
Unrecognized keyword when using 'x' as table alias


svivian:
Hopefully this is the right place for this bug. In phpMyAdmin I get an error when using a legitimate query. Using the "Country" table on the demo server as an example, putting the following query into the SQL box gives an error:

```sql
update Country x, City c
set x.Name=x.Name
where x.Code=c.CountryCode;
```

The first x gets underlined in red and it says "Unrecognized keyword (near x)". But the query is valid and submitting it runs just fine.

I don't get the error in a select query (e.g. `select * from Country x, City c...`) and changing the x to another letter works fine (e.g. `Country w, City c`).

williamdes:
Hi @svivian
This is indeed the right place to report such bugs
Thank you

niconoe-:
@williamdes  / @iifawzi 
I wanted to give it a try on this issue, but I have the feeling that using SQL-Parser 5.7.0 (and thanks to #385), this bug is already fixed. I just tried with 
```
php bin/highlight-query --query "update Country x, City c
set x.Name=x.Name
where x.Code=c.CountryCode;"
```

and I got a successful answer
```
UPDATE
    Country x,
    City c
SET
    x.Name = x.Name
WHERE
    x.Code = c.CountryCode;
```

wihout any error.

Are you aware of such statement that could reproduce the issue?
Thanks a lot!

williamdes:
Indeed, I think it was a mistake and I forgot to close the issues :O

But that is also due to the fact that `Fixes #xyz` should have been the wording that github would detect and close the issues on PR merge 😸

williamdes:
> @williamdes / @iifawzi I wanted to give it a try on this issue, but I have the feeling that using SQL-Parser 5.7.0 (and thanks to #385), this bug is already fixed. I just tried with
> 
> ```
> php bin/highlight-query --query "update Country x, City c
> set x.Name=x.Name
> where x.Code=c.CountryCode;"
> ```
> 
> and I got a successful answer
> 
> ```sql
> UPDATE
>     Country x,
>     City c
> SET
>     x.Name = x.Name
> WHERE
>     x.Code = c.CountryCode;
> ```
> 
> wihout any error.
> 
> Are you aware of such statement that could reproduce the issue? Thanks a lot!

Probably the one one on #327

niconoe-:
> Probably the one on #327

Tested and working as expected:
``` 
[...]

Issue #327, opened 2021-01-29
Parser does not recognize single character table name or alias


joseffi:
When using a single character as a table alias or name, the table.column names are not properly recognized

For example, this SQL:

```sql
SELECT x.asset_id FROM
    (SELECT evt.asset_id FROM evt) x
```
is incorrectly parsed into:

```            [expr] => Array
                (
                    [0] => PhpMyAdmin\SqlParser\Components\Expression Object
                        (
                            [database] => 
                            [table] => 
                            [column] => 
                            [expr] => x.asset_id
                            [alias] => 
                            [function] => 
                            [subquery] => 
                        )

                )
```

But this works:
```sql
SELECT xx.asset_id FROM
    (SELECT evt.asset_id FROM evt) xx
```

parser result: 
```            [expr] => Array
                (
                    [0] => PhpMyAdmin\SqlParser\Components\Expression Object
                        (
                            [database] => 
                            [table] => xx
                            [column] => asset_id
                            [expr] => xx.asset_id
                            [alias] => 
                            [function] => 
                            [subquery] => 
                        )

                )
```

williamdes:
Thank you for reporting this issue !

williamdes:
Fixed by #385

PR #385, opened 2022-08-19
Fixing parser not recognizing single chars aliases or tables

iifawzi:
Fixing: #327  and #378 


iifawzi:
Regarding phpstan, I think we would need to update the baseline, what do you think @williamdes?

src/Components/Expression.php
```diff
@@ -225,7 +225,7 @@ public static function parse(Parser $parser, TokensList $list, array $options =
                 continue;
             }
 
-            if ($token->type === Token::TYPE_KEYWORD) {
+            if ($token->type === Token::TYPE_KEYWORD && ($token->value !== 'x' && $token->value !== 'y')) {
```
williamdes, line 228: I am not sure this is correct, maybe we should add a constant array in this file or change something elsewhere

iifawzi: I'd go with the constant array, but what would be an appropriate name for it? 
I don't think that there will be a more appropriate place to change this, because we don't need to mess around with the how the lexer is behaving or how the `X` or `Y` is parsed elsewhere other than here.

williamdes: Can we detect that it is a function ?
I really do not like that we hard code two cases that could bring more afterwards

iifawzi: There's no part of the entire project knows that those are functions, the issue is that the lexer is tokenizing them as `keywords`.

If we want the parser to be aware that those are functions, maybe we can modify the lexer to have a boolean property in the tokens dictionaries, `function` for example, or something, that can indicate whether it's a function or not, and this can be done I think because the functions have (F) appended in the end, at the contexts files. 

This way, we would be able to know anywhere in the code whether it's a function or not. 
what do you think @williamdes

williamdes: > and this can be done I think because the functions have (F) appended in the end, at the contexts files.

Indeed, this is probably our starting point

> This way, we would be able to know anywhere in the code whether it's a function or not.
> what do you think @williamdes

I re-read this today and I think it is a viable and good way of doing this. It will make us to be able to lint functions in a better way too.

iifawzi: Perfect, I will make sure to find time to work on this

williamdes: Thank you, that's very nice

williamdes:
Note: When this gets merged please hard reset your master branch to phpmyadmin/sql-parser#master and use specific branch names for your PRs ;)

iifawzi:
Yes, i've mistakenly used the master branch here, I will make sure to do this after this get merged

src/Lexer.php
```diff
@@ -385,6 +386,49 @@ private function solveAmbiguityOnStarOperator()
...
+                continue;
+            }
+
+            $keywordToken->type = Token::TYPE_NONE;
+            $keywordToken->flags = Token::TYPE_NONE;
+            $keywordToken->keyword = $keywordToken->value;
+        }
+
+        $this->list->idx = $iBak;
+    }
```
iifawzi, line 430: I hope that I'm not missing any other cases. This way the keywords that represent functions names can be used freely as aliases or table names

iifawzi (commented):
changes

src/Lexer.php
```diff
@@ -385,6 +386,49 @@ private function solveAmbiguityOnStarOperator()
...
+     * - "." (a dot separator like in "x.asset_id FROM (SELECT evt.asset_id FROM evt)".
+     * - "NULL" (when used as a table alias like in "avg.col FROM (SELECT ev.col FROM ev) avg").
+     *
+     * This method will change the flag of the function keyword tokens when any of those
+     * condition above is true. Otherwise, the
+     * default flag (function keyword) will be kept.
+     *
+     * @return void
+     */
+    private function solveAmbiguityOnFunctionKeywords()
```
williamdes, line 410:
```suggestion
     */
    private function solveAmbiguityOnFunctionKeywords(): void
```

src/TokensList.php
```diff
@@ -168,6 +168,25 @@ public function getNextOfTypeAndValue($type, $value)
...
 
+    /**
+     * Gets the next token.
+     *
+     * @param int $type the type of the token
+     * @param int $flag the flag of the token
+     *
+     * @return Token|null
+     */
+    public function getNextOfTypeAndFlag($type, $flag)
```
williamdes, line 179: I think this can get type hints too

src/Lexer.php
```diff
@@ -385,6 +386,49 @@ private function solveAmbiguityOnStarOperator()
...
+    {
+        $iBak = $this->list->idx;
+        $keywordFunction = Token::TYPE_KEYWORD | Token::FLAG_KEYWORD_FUNCTION;
+        while (($keywordToken = $this->list->getNextOfTypeAndFlag(Token::TYPE_KEYWORD, $keywordFunction)) !== null) {
+            $next = $this->list->getNext();
+            if (
+                ($next->type !== Token::TYPE_KEYWORD || ! in_array($next->value, ['FROM', 'SET', 'WHERE'], true))
+                && ($next->type !== Token::TYPE_OPERATOR || ! in_array($next->value, ['.', ','], true))
```
williamdes, line 418: the two arrays should be constants of this class for more clarity, right ?

iifawzi: I don't think so, we're not using them anywhere else other than here, what do you think?

williamdes: I think the parser uses properties that look like consts and maybe we should continue doing like that. And it makes it more clear :)

williamdes:
Can you squash your commits on the next push ?

williamdes (approved):
This looks really nice, I think we should wait for the parser to be released before merging this into it. I consider this one as potentially risky, not sure why.
But anyway we can agree this is not emergency and can wait for the next window to go in ?

iifawzi:
Since it affects the lexer, I agree that it's potentially risky. But still, the risks that this might introduce are minimal since we're only changing the token type in very specific cases, and otherwise it's kept as it's.

Anyway, this's not an emergency yes, it can for sure wait for the next release.

PR #386, opened 2022-11-12
Differentiating between analyze and explain statements

iifawzi:

Hi, this should fix https://github.com/phpmyadmin/phpmyadmin/issues/17482#issuecomment-1312239801
>seems that ANALYZE maps to AnalyzeStatement, which is seen as "Table Maintenance Statements", which is fine, for ANALYZE TABLE.

>However, ANALYZE SELECT is similar to EXPLAIN (in MariaDB), so it probably should be seen as ExplainStatement.

Ref: https://github.com/phpmyadmin/phpmyadmin/issues/17482#issuecomment-1312471693.

We still need to implement the `EXPLAIN` statement, but this might take place in a separate PR.


src/Parser.php
```diff
@@ -466,27 +466,44 @@ public function parse()
...
+            ++$list->idx; // Skip ANALYZE
+            $first = $list->getNextOfType(Token::TYPE_KEYWORD);
+            $second = $list->getNextOfType(Token::TYPE_KEYWORD);
+            $statementName = null;
+
+            // ANALYZE keyword can be an indication of two cases:
+            // 1- ANALAYZE TABLE statements, in both mariaDB and MySQL
+            // 2- Explain statement, in case of mariaDB https://mariadb.com/kb/en/explain-analyze/
+            // We need to point case 2 to use the EXPLAIN Parser.
+            if ($token->keyword === 'ANALYZE' && $first->keyword !== 'TABLE' && $second->keyword !== 'TABLE') {
```
iifawzi, line 480: as per https://dev.mysql.com/doc/refman/5.6/en/analyze-table.html and https://mariadb.com/kb/en/analyze-table/ANAL. TABLE can only be the first or the second keyword token after analyze. by checking this, we are able to determine whether it's a table maintenance statement or not

src/Parser.php
```diff
@@ -466,27 +466,44 @@ public function parse()
...
-            if (empty(static::$STATEMENT_PARSERS[$token->keyword])) {
-                if (! isset(static::$STATEMENT_PARSERS[$token->keyword])) {
-                    // A statement is considered recognized if the parser
-                    // is aware that it is a statement, but it does not have
-                    // a parser for it yet.
-                    $this->error('Unrecognized statement type.', $token);
-                }
+            $lastIdx = $list->idx;
+
+            ++$list->idx; // Skip ANALYZE
```
williamdes, line 471: how do we know for sure we are skipping analyse ?

src/Parser.php
williamdes, line 480: should that be in another file specific to select ?

src/Parser.php
iifawzi: In fact, we don't, the comment is misleading. 

We can have an if condition on the keyword, and if it's ANALYZE we can have all the checks nested, maybe: 

```PHP
            $lastIdx = $list->idx;

            if ($token->keyword === 'ANALYZE') {

                ++$list->idx; // Skip ANALYZE
                $first = $list->getNextOfType(Token::TYPE_KEYWORD);
                $second = $list->getNextOfType(Token::TYPE_KEYWORD);
                $statementName = null;
                // ANALYZE keyword can be an indication of two cases:
                // 1- ANALAYZE TABLE statements, in both mariaDB and MySQL
                // 2- Explain statement, in case of mariaDB https://mariadb.com/kb/en/explain-analyze/
                // We need to point case 2 to use the EXPLAIN Parser.
                if ($first->keyword !== 'TABLE' && $second->keyword !== 'TABLE') {
                    $statementName = 'EXPLAIN';
                } else {
                    $statementName = 'ANALYZE';
                }
                $list->idx = $lastIdx;

            } else {
                $list->idx = $lastIdx;
                // Checking if it is a known statement that can be parsed.
                if (empty(static::$STATEMENT_PARSERS[$token->keyword])) {
                    if (! isset(static::$STATEMENT_PARSERS[$token->keyword])) {
                        // A statement is considered recognized if the parser
                        // is aware that it is a statement, but it does not have
                        // a parser for it yet.
                        $this->error('Unrecognized statement type.', $token);
                    }

                    // Skipping to the end of this statement.
                    $list->getNextOfType(Token::TYPE_DELIMITER);
                    $prevLastIdx = $list->idx;
                    continue;
                }
            }
```

and this way, we can use the same logic to handle any special case similar to ANALYZE in the future, and for normal statements it will reach the else block.

what do you think? @williamdes.

src/Parser.php
iifawzi: didn't get what you mean by specific to select, can you elaborate more?

src/Parser.php
iifawzi: The suggestion above will complicate the statement selection. I will take another look.

williamdes: I think your suggestion is cleaner and clearer. It shows it's a temporary workaround until it is implemented.
But for what this is worth can we not create a statement file for this one and really implement it in a light way that can be improved ?

like just creating the statement so the workaround code is not needed and then in the statement file have the minimum for now ?

iifawzi: This checks here are not because it's not implemented yet. it's because we have two type of queries that might start with `ANALYZE`: 

1- `ANALYZE TABLE .....`
2- `ANALYZE SELECT.....`

for type 1, we want to use the `ANALYZE` parser. 
for type 2, we want to use the `EXPLAIN` parser. 

so the checks here are just to route the statement to the correct parser class.

https://github.com/phpmyadmin/sql-parser/blob/a9cd167be77829ec2dd4496ce1010ed6abde92f8/src/Parser.php#L28-L42


so even if we implemented the EXPLAIN parser, we will still need this routing. 
what do you think? I might be missing something

iifawzi: The mess came from that MariaDB is using `ANALYZE` instead of `EXPLAIN ANALYZE` in the new versions.

williamdes: Okay, so I think the suggested code above can be used for this PR then (I made some edits to it)

iifawzi: Sorry, just noticed your changes. 
Committed

src/Parser.php
```diff
@@ -466,27 +466,48 @@ public function parse()
...
+                }
             }
 
             /**
              * The name of the class that is used for parsing.
              *
              * @var string
              */
-            $class = static::$STATEMENT_PARSERS[$token->keyword];
+            $class = static::$STATEMENT_PARSERS[$statementName ? $statementName : $token->keyword];
```
williamdes, line 510:
```suggestion
            $class = static::$STATEMENT_PARSERS[$statementName !== null ? $statementName : $token->keyword];
```

nunoperalta:
@iifawzi  - thank you very much for this!

I saw your test cases,

can you please also add another test?

`ANALYZE FORMAT=JSON SELECT * FROM orders`

PR #394, opened 2022-11-27
Testing the `contextGenerator` class

iifawzi:


iifawzi:
can't find a proper name to test this? could you help @williamdes 
https://github.com/phpmyadmin/sql-parser/blob/a70dec025bbcf9bddfc42117a29226b94694cf30/src/Tools/ContextGenerator.php#L292-L294

iifawzi:
Most probably, this's a dead code
https://github.com/phpmyadmin/sql-parser/blob/a70dec025bbcf9bddfc42117a29226b94694cf30/src/Tools/ContextGenerator.php#L177-L180

it checks whether the keyword is of length zero? how could this be possible?

tests/Tools/contexts/testContext.txt
```diff
@@ -0,0 +1,14 @@
+RESERVED (R)
+RESERVED2 (R)
+RESERVED3 (R)
+RESERVED4 (R)
+RESERVED5 (R)
```
iifawzi, line 5: Multiple keywords are added here to test that the formatted will use new line

williamdes:
Hi,

Thank you for this, I will have a look when I finish my refactoring upstream at mariadb-mysql-kbs. I am near to have re-implementing the scraping into Rust code. Next should follow data for keywords and if possible functions and everything else.

I was thinking it would be nice to:
- reduce the format of context files to a minimum with sections like:
    - Added
    - Removed
    - Changed

 When using a vscode plugin to have the diff between files I did see that most of the file remains the same
It makes it uneasy to track changes and see missing elements.
We should rather build a diff system because it's real world about what is exactly the case.

Common is version v0 and MariaDB is v1 and so on and a version is nothing without the previous one. So it's like git.
If the contexts end up cleaned up like that we can fire up my upstream changes and ask for version x.y.z what are the known keywords and compare the lists to find missing on wrong ones.
Ideally when upstream and the parser are reconciled we can (maybe?) delete the context files and use my lib as source of truth. And still make manual updates of contexts and keywords

williamdes (approved):

PR #397, opened 2022-12-07
Performance improvement to use less the "nextToken()" function

iifawzi:
Hi, This's an improvement to skip the current token by increasing the `idx`, instead of the additional `nextToken` call. 


williamdes (approved):
🚀

PR #398, opened 2022-12-07
Fix #396 - Supporting `is null` and `is not null` with order by

### LINKED ISSUE #396: Parser does not allow IS NULL in ORDER BY
The following query leads to errors:
```sql
SELECT * FROM some_table
ORDER BY some_col IS NULL;
```
There is a workaround, by using parentheses:
```sql
SELECT * FROM some_table
ORDER BY (some_col IS NULL):
```
However, I think the first query should be allowed, as it is valid MySQL.

iifawzi:
Hi, this PR should fix https://github.com/phpmyadmin/sql-parser/issues/396

Fixes https://github.com/phpmyadmin/sql-parser/issues/396


williamdes (approved):
💯 Thank you!

iifawzi:
This should be closed in favor of #399

Issue #171, opened 2017-09-08
Use of certain keywords in SELECT field list


stevenwdv:
Some queries with keywords in the `SELECT` field list are marked as invalid by the SQL parser.
Examples:
- `SELECT NULL IS NULL`
- `SELECT NOT 1`
- `SELECT 1 BETWEEN 0 AND 2`

Wrapping the expression in parentheses avoids the errors, eg. `SELECT (NULL IS NULL)`.

**phpMyAdmin version:** 4.7.4

stevenwdv:
This also happens with `SELECT NOT 1`

nijel:
Reproduced with current master:

```
./bin/lint-query --query 'SELECT NULL IS NULL'
#1: Unrecognized keyword. (near "IS" at position 12)
#2: Unrecognized keyword. (near "NULL" at position 15)
./bin/lint-query --query 'SELECT NOT 1'
#1: An expression was expected. (near "NOT" at position 7)
#2: Unrecognized keyword. (near "NOT" at position 7)
#3: Unexpected token. (near "1" at position 11)
./bin/lint-query --query 'SELECT 1 BETWEEN 0 AND 2'
#1: Unrecognized keyword. (near "BETWEEN" at position 9)
#2: Unexpected token. (near "0" at position 17)
#3: Unrecognized keyword. (near "AND" at position 19)
#4: Unexpected token. (near "2" at position 23)
```

Not sure if those are not three separate issues...

PR #399, opened 2022-12-12
Improving the `Expression` component's allowed keywords

iifawzi:
Hi, This PR is improving the `Expression` component allowed keywords

Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/17422
Fixes: https://github.com/phpmyadmin/sql-parser/issues/396
Fixes: https://github.com/phpmyadmin/sql-parser/issues/171


williamdes (approved):
💯

tests/data/parser/parseCreateViewWithoutQuotes.out
```diff
@@ -1269,17 +1276,20 @@
...
+                                    "OVER": 2,
+                                    "IS": 3,
+                                    "NOT NULL": 4,
+                                    "NULL": 4
                                 },
                                 "database": null,
                                 "table": null,
                                 "column": "root",
-                                "expr": "root@localhost",
+                                "expr": "root @localhost",
```
williamdes, line 1287: this one is not good

src/Components/Expression.php
```diff
@@ -371,6 +375,19 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
                 } else {
+                    $currIdx = $list->idx;
+                    --$list->idx;
+                    $beforeToken = $list->getPrevious();
+                    $list->idx = $currIdx;
+                    if (
+                        $ret->expr !== null && $beforeToken &&
+                        $beforeToken->type !== Token::TYPE_OPERATOR &&
+                        $token->type !== Token::TYPE_OPERATOR &&
+                        $token->type !== Token::TYPE_SYMBOL
```
williamdes, line 386: would it not be simpler to only state the token that can be used to enter this if ?

williamdes (approved):

src/Components/Expression.php
iifawzi: Yes it would, and will help decreasing the unnecessarily spaces too, I will work on the changes

iifawzi:
We need to update all the out files to contain the new added keywords, should this be added on a separate PR?

williamdes (approved):
That's awesome and minimal miel we love it

williamdes:
> We need to update all the out files to contain the newly added options, should this be added on a separate PR?

I would say yes, but do not worry I will do it
I fill avoid more un necessary work on your end

src/Components/Expression.php
```diff
@@ -371,6 +375,23 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
+                    $beforeToken = $list->getPrevious();
+                    $list->idx = $currIdx;
+                    // columns names tokens are of type NONE, or SYMBOL (`col`), and the columns options
+                    // would start with a token of type KEYWORD, in that case, we want to have a space
+                    // between the tokens.
+                    if (
+                        $ret->expr !== null &&
+                        $beforeToken &&
+                        ($beforeToken->type === Token::TYPE_NONE ||
+                        $beforeToken->type === Token::TYPE_SYMBOL) &&
```
iifawzi, line 389: a tiny improvement to handle the case if the column is surrounded with backticks, the col token then is considered  `SYMBOL`, so added a check on that type too.

williamdes: Cool, is that covered by the unit test ?

iifawzi: Yes, I've updated the tests
```PHP
        $parser = new Parser('SELECT `test3`.`t1` is not null AS `is_not_null` FROM `test3` ;');
        $stmt = $parser->statements[0];

        $this->assertEquals('SELECT `test3`.`t1` is not null AS `is_not_null` FROM `test3`', $stmt->build());
```

williamdes: 🚀 
Ready for merge then, I will have a look today

src/Components/Expression.php
```diff
@@ -31,19 +32,21 @@ class Expression extends Component
...
         'DIV' => 1,
-        'AND' => 1,
+        'IS' => 1,
+        'MOD' => 1,
+        'NOT' => 1,
+        'NOT NULL' => 1,
+        'NULL' => 1,
         'OR' => 1,
+        'OVER' => 1,
+        'REGEXP' => 1,
```
williamdes, line 48: maybe also support NOT REGEXP and `RLIKE` https://dev.mysql.com/doc/refman/8.0/en/regexp.html

iifawzi: NOT REGEXP is already supported because NOT and REGEXP are separate keywords and both are supported. 
I've added the support of RLIKE, and improved the space handling, to be able to have space if the before the token is a string, like in the case of 
```PHP
SELECT 'Michael!' REGEXP '.*';
```

williamdes: Awesome !!

PR #408, opened 2023-01-09
Fix #404 - Implement `ALTER EVENT`

### LINKED ISSUE #404: "ALTER EVENT" is not recognized by the SQL parser
It seems that the latest release of sql-parser (5.6.0) does not support "ALTER EVENTS" yet.
The AlterOperation component needs to be updated in order to add all the options for the EVENT.
Any plans to handle this kind of statements please?

niconoe-:
Hello all.

I'm working IRL with @SanaRebhi and we tried to make a proposal in order to solve this topic.

I made some manual tests and it looks like it's OK, but I want to go deeper and add unit tests also to cover (almost) all test cases about ALTER EVENT statements, which I don't have time right now.

Current unit-test suite is passing (with 2 ignored test-cases out of this range) locally.

Please feel free to tell me if this approch is a good one and I can continue to write tests or if you have another way in mind to manage it.

Thanks a lot.

Fixes: #404


williamdes:
I think @iifawzi would better know if this feels right or not
But from what I see it's nice

src/Contexts/ContextMySql80000.php
```diff
@@ -180,6 +180,8 @@ class ContextMySql80000 extends Context
         'DEFAULT CHARACTER SET' => 7,
         'NATURAL LEFT OUTER JOIN' => 7,
         'NATURAL RIGHT OUTER JOIN' => 7, 'WITH CONSISTENT SNAPSHOT' => 7,
+        'ON COMPLETION PRESERVE' => 7, 'ON COMPLETION NOT PRESERVE' => 7,
```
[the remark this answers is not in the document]
niconoe-: Thanks. I will do it and update my PR soon.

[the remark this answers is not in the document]
niconoe-: @williamdes What means the `(R)`, `(D)`, `(F)` or `(K)` at the end of some lines please?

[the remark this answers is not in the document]
williamdes: This should explain it
https://github.com/phpmyadmin/sql-parser/blob/c8a83aaece057d0214190595992d1e69f0458c5a/src/Tools/ContextGenerator.php#L45-L50

tools/contexts/_common.txt
```diff
@@ -20,6 +20,7 @@ DECIMAL (D)
 DEFAULT CHARACTER SET
 DEFAULT CHARSET
 DEFAULT COLLATE
+DISABLE ON SLAVE
```
williamdes, line 23: So all keywords you added exist on all MySQL and MariaDB versions ?

If not you must add them in the first version/context file that it was supported :)

niconoe-: I really thought it was, but your comment made me double check and I misread the documentation of MySQL 5.0 (I though I was on the doc of MySQL 5.5)

I'll update this and double check every version. Sorry about it.

williamdes: No worries
I wanted to be sure you has understood how the context files work :)
And you need to add on all of them up to the latest version (not sure but maybe ^^)

niconoe-: MySQL EVENTS are available since MySQL 5.1.x but I can't find the `x` value and I don't know if this is starting from `5.1.0` or a little more recent version… As MariaDB 10.0 comes from MySQL 5.5 and the EVENT statements are not changed from MariaDb to MySQL (nor opposite), I'll put those keywords into all MariaDB context files 

About MySQL 5.1, I don't know to be honest as I can't find any documentation related to only MySQL 5.1.0 specifically: I only have the documentation of the latest MySQL 5.1.

Therefore, I'll add them to all contexts but MySQL 5.0 and MySQL 5.1

niconoe-: @williamdes Maybe you didn't get notified by my question.

Knowing that events are a feature available only since MySQL 5.1.6, so keywords like `DISABLE ON SLAVE` and so on too, should I add the keywords in the `MySql50100.txt` file (even if those keywords are not available in MySql 5.1.0)?

Thanks.

williamdes: I had the answer in my emails but your last reply is easier to reply and understand
I would say yes, you can add them
One day we will improve this but it's okay

williamdes:
Would you mind opening a new pull-request with all of your commits but not the two source file changes ?
Or a new one with only the two source file changes ?

This is because I would like to merge all the testing part so we can see the difference after the implementation is made and what it was before in terms of `.out` files.
What I always do for such PRs is
```sh
rm tests/data/**/*.out
./tools/run_generators.sh
```
To have the changes after a new implementation on all out files

niconoe-:
> Would you mind opening a new pull-request with all of your commits but not the two source file changes ? Or a new one with only the two source file changes ?
> 
> This is because I would like to merge all the testing part so we can see the difference after the implementation is made and what it was before in terms of `.out` files. What I always do for such PRs is
> 
> ```shell
> rm tests/data/**/*.out
> ./tools/run_generators.sh
> ```
> 
> To have the changes after a new implementation on all out files

I’m not sure I get it.
If I send a PR without the source changes, the unit test won’t pass anymore as the .out files and the parser’s results will differ.
If I send a PR with only the source changes, the coverage report won’t like it.


I guess what you want is the list of .in files and the .out files but generated from a previous state (when the source files weren’t changed) so that with this current PR (#408), you’ll see only the source files changes + updates of the .out files, and you would be capable of comparing those .out files, am I right?

I’ll do this today: providing a PR with new keywords, new unit tests about ALTER EVENT statements, but without the feature implementation. Is this ok for you?

williamdes:
> I guess what you want is the list of .in files and the .out files but generated from a previous state (when the source files weren’t changed) so that with this current PR (#408), you’ll see only the source files changes + updates of the .out files, and you would be capable of comparing those .out files, am I right?

Yes exactly 
The in files generated with the current code in the master branch and the context files added

And yes the goal is to see the diff in out files with the new implementation, you got it right 

> I’ll do this today: providing a PR with new keywords, new unit tests about ALTER EVENT statements, but without the feature implementation. Is this ok for you?

Yes

niconoe-:
@williamdes Here is the PR: #409. Please tell me if something is wrong, I'll try to fix it ASAP.

Thanks again.

src/Components/AlterOperation.php
```diff
@@ -299,7 +340,9 @@ public static function parse(Parser $parser, TokensList $list, array $options =
             if ($state === 0) {
                 $ret->options = OptionsArray::parse($parser, $list, $options);
 
-                if ($ret->options->has('AS')) {
+                // Not only when aliasing but also when parsing the body of an event, we just list the tokens of the
+                // body in the unknown tokens list, as they define their own statements.
+                if ($ret->options->has('AS') || (isset($options['EVENT']) && $ret->options->has('DO'))) {
```
iifawzi, line 345: do we really need to have `EVENT` in the options? can't we just check the existence of `DO`?

src/Statements/AlterStatement.php
```diff
@@ -115,6 +116,8 @@ public function parse(Parser $parser, TokensList $list)
                     $options = AlterOperation::$VIEW_OPTIONS;
                 } elseif ($this->options->has('USER')) {
                     $options = AlterOperation::$USER_OPTIONS;
+                } elseif ($this->options->has('EVENT')) {
+                    $options = array_merge(AlterOperation::$EVENT_OPTIONS, ['EVENT' => 0]);
```
iifawzi, line 120: I think we need to avoid doing this. it's kinda illogical from the statement point of view to have `EVENT` as an option. Most probably as per what I understand, you made it this way to be able to know whether it's an `ALTER EVENT` statement or not, right? so you can add the tokens after `DO` to the unknowns

if we can't just depend on `DO`, we need to see if there's a better approach, wdyt?

iifawzi (changes requested):

williamdes:
> @williamdes Here is the PR: #409. Please tell me if something is wrong, I'll try to fix it ASAP.
> 
> Thanks again.

I merged #409 thank you for this effort 🎉 
You can rebase and re-generate the out files :)

iifawzi:
> I added some unit tests, I hope I did it well.
> 
> AFAIK, this is now ready for review.
> 
> Just to let you know, there are just 1 drawback I have found: When you have an `ALTER EVENT` statement with dissociative options (like `ENABLE` and `DISABLE` at the same time), those options are not entering into conflict even if their IDs are declared as the same. As this behavior is generic to all parsing options but not only on `ALTER EVENT`, I prefered to ignore this and try to understand why in a next issue/PR.

This's an issue that impacts many of the components, you're right. only the `optionsArray` component has the correct checks, this's definitely an area for improvement for the rest of the components

src/Components/AlterOperation.php
niconoe-: I was wondering this too, but the fact is during an ALTER EVENT statement, you might not need to rewrite your event. For instance `ALTER EVENT my_event ENABLE;` is possible, and yet, you don't have any `DO` keyword.

src/Statements/AlterStatement.php
niconoe-: We can't depend on the `DO` as it's not a mandatory keyword during an `ALTER EVENT`. 

I didn't find another way to make it sure options are parsed in a context of an `ALTER EVENT` statement as, in `OptionArray` class, there's absolutly no way to know which kind of statement the options we are parsing are comming. This `EVENT` fake-option is the only possible flag I could found, even after trying to find a better approach :(.

iifawzi: if event body is not present, ( `DO` is not used ), we don't have any unknown tokens I think, wdyt? this way we don't need to check the existence of `EVENT`, we can only depend on `DO`. 

what I mean is that, by also looking at your tests output files, no unknown tokens are found except for the case when the event definition is present ( parseAlterEventComplete, parseAlterEvent8 )

src/Components/AlterOperation.php
niconoe-: Also, I didn't want to impact any other possible statement that may have a `DO` with a list of tokens to parse. If the `DO` is only used in this way (like `DO <final list of tokens and nothing else>`), well, we could remove the "EVENT" check. But I'm not enough experienced with MySQL and MariaDB to know if we could have another kind of statement where the `DO` can have another meaning.

niconoe-:
Failing test about mutation is about network issue I would say. Would it be possible to re-run it please?

tests/data/parser/parseAlterEvent8.out
```diff
@@ -667,40 +667,74 @@
...
+                            {
+                                "@type": "@14"
+                            },
+                            {
+                                "@type": "@15"
+                            },
+                            {
+                                "@type": "@16"
+                            }
+                        ]
```
williamdes, line 737: I think there is some missing code to handle this case fully

tests/data/parser/parseAlterEventComplete.out
```diff
@@ -1081,114 +1081,187 @@
...
-                            },
-                            {
-                                "@type": "@39"
-                            },
-                            {
-                                "@type": "@40"
-                            },
-                            {
-                                "@type": "@41"
+                                "@type": "@62"
```
williamdes, line 1264: this one too

src/Statements/AlterStatement.php
niconoe-: Ok, thanks. I'll try to analyse this without the `EVENT` keyword check, but only the `DO`. Maybe this will be ok. Thanks for the feedback

tests/data/parser/parseAlterEvent8.out
niconoe-: I don't get what's expected here. 

For instance, if you take a look at `parseAlterErr.out` (which is not related to ALTER EVENT), you will have a list of unknown tokens just like this.

In the case of ALTER EVENT, the tokens in the `DO` statements are considered as `unknown` since there's no options that can handle a TokenList in the properties of the AlterStatement class. That's the reason of the change in `AlterOperation.php`.

In this case, the .in file is 
```sql
ALTER EVENT my_event DO
  SELECT * FROM my_table;
```
So I expect 8 tokens:
1) `\n  `
2) `SELECT`
3) ` `
4) `*`
5) ` `
6) `FROM`
7) ` `
8) `my_table`

And that's exactly what is expected by the .out file here.

src/Statements/AlterStatement.php
niconoe-: I tried without the `EVENT` flag, and you were right, the unit tests still passes. This was useless so I will remove it.

Thanks a lot @iifawzi !

tests/data/parser/parseAlterEvent8.out
niconoe-: Yup, probably would be better. But in the meantime, either we can let the tokens of the `DO` in the `unknown`, or I can provide a new AlterOperation property just for that. I dislike the idea of having a dedicated property as I have the feeling it will not respect the Single Responsibility rule, or at least increase the number of responsibilities the AlpterOperation class has, and I didn't want to add code smells here.

As you wish, I can do it anyway if you prefer to expect 0 tokens on the `unknown` array.

iifawzi: We don't really need to add a new `alterOperation` handler for the `DO` statements,  instead, it would be better if encapsulated all the logic of handling the `DO` statement in a separate class, similar to what we're doing with all the statements. we will be able then to remove `DO` from the `EVENT_OPTIONS` and delegate the handling of the rest of the keywords that follows DO to the `DoStatement` parse class as we're doing here: 

https://github.com/phpmyadmin/sql-parser/blob/9306568dd9ff798fb50ef38f1cf6b93ff046bd63/src/Statements/ExplainStatement.php#L165-L181

and here https://github.com/phpmyadmin/sql-parser/blob/9306568dd9ff798fb50ef38f1cf6b93ff046bd63/src/Statements/WithStatement.php#L242-L257

src/Statements/AlterStatement.php
iifawzi: Thanks to you Nicolas, you've done very great work here :')

tests/data/parser/parseAlterEvent8.out
niconoe-: I think I got your point, and I agree with you. Nevertheless, I have the feeling that creating the `DoStatement` class with all required logic + implementing it for the ALTER EVENT, is a little bit too huge for my small shoulders :smile: .

IMO, it could be a very interesting feature to come, but is not necessary yet to allow `ALTER EVENT` statements to be understood by the Parser. With the actual version 5.6.0, when you ask for `ALTER EVENT`, you just have an error saying that this statement is not recognized at all, and at least, this won't be the case anymore with my PR, even if I admit some improvements cas be done.

williamdes: > I think I got your point, and I agree with you. Nevertheless, I have the feeling that creating the DoStatement class with all required logic + implementing it for the ALTER EVENT, is a little bit too huge for my small shoulders smile .

Thats's okay for me
If @iifawzi approves this PR it's ready for a merge then ;p

williamdes (approved):

tests/data/parser/parseAlterEvent8.out
iifawzi: yes, It looks good to me, we can implement to `DO` statement later, and integrate it with this. 

If I can approve I would do so, but I can't :p.

niconoe-: All threads are resolved and checks are passing. Do you know what prevents you to approve?

iifawzi (approved):

tests/data/parser/parseAlterEvent8.out
iifawzi: > All threads are resolved and checks are passing. Do you know what prevents you to approve?

I thought that I'm not able to approve because I'm not a member, anyway, it worked now, approved. 

Thank you @niconoe- for this precious work

niconoe-:
> Merci for your patience and hard work @niconoe- !

De rien, et avec plaisir ;)

PR #411, opened 2023-01-15
Fix #406 - Fixing createStatement function's options

### LINKED ISSUE #406: Exporting procedures with 'MODIFIES SQL DATA' generates a warning in OptionsArray.php
### Describe the bug

When you export a database with the 'Add CREATE PROCEDURE / FUNCTION / EVENT statement' option enabled, and the procedure definition includes 'MODIFIES SQL DATA', the parser generates the following message:

```
Warning in ./vendor/phpmyadmin/sql-parser/src/Components/OptionsArray.php#306
Undefined array key "value"
```

Dumping the value of $option just before that shows the following as the cause of the message:

```
Array
(   
    [name] => MODIFIES
    [equals] =>
    [expr] =>
)
```

### To Reproduce

Create a procedure with 'MODIFIES SQL DATA' as the SQL data access option.
Click the Export tab for the database with said procedure.
Perform a custom export with the 'Add CREATE PROCEDURE / FUNCTION / EVENT statement'
See the warning and backtrace.

### Expected behavior

No warning messages during the export.

### Screenshots

Here's the full backtrace:
```
Warning in ./vendor/phpmyadmin/sql-parser/src/Components/OptionsArray.php#306
Undefined array key "value"

Backtrace

CreateStatement.php#491: PhpMyAdmin\SqlParser\Components\OptionsArray::build()
ExportSql.php#2814: PhpMyAdmin\SqlParser\Statements\CreateStatement->build()
ExportSql.php#558: PhpMyAdmin\Plugins\Export\ExportSql->replaceWithAliases(
string 'CREATE DEFINER=`root`@`localhost` PROCEDURE `split_string`(IN `x` VARCHAR(255), IN `delim` VARCHAR(12), IN `id` CHAR(36)) MODIFIES SQL DATA DETERMINISTIC BEGIN SET @valcount = substrCount(x,delim)+1; SET @vl = 0; CREATE TEMPORARY TABLE IF NOT EXISTS splitResultsTemp (uuid char(36), split_value varchar(255)) ENGINE=MEMORY; WHILE (@vl < @valcount) DO SET @val = stringSplit(x,delim,@vl+1); INSERT INTO splitResultsTemp (uuid, split_value) VALUES (id, @val); SET @vl = @vl + 1; END WHILE; END',
array,
string 'ftp_accounts',
string '',
boolean false,
)
ExportSql.php#620: PhpMyAdmin\Plugins\Export\ExportSql->exportRoutineSQL(
string 'ftp_accounts',
array,
string 'PROCEDURE',
string 'Procedures',
array,
string '$$',
)
Export.php#742: PhpMyAdmin\Plugins\Export\ExportSql->exportRoutines(
string 'ftp_accounts',
array,
)
ExportController.php#548: PhpMyAdmin\Export->exportDatabase(
string 'ftp_accounts',
array,
string 'structure_and_data',
array,
array,
,
string ' ',
string 'index.php?route=/database/export&db=ftp_accounts',
string 'database',
boolean true,
boolean true,
boolean true,
boolean true,
array,
string '',
)
Routing.php#192: PhpMyAdmin\Controllers\Export\ExportController->__invoke(
,
array,
)
index.php#43: PhpMyAdmin\Routing::callControllerForRoute(
,
string '/export',
,
,
)
```

### Server configuration

- Operating system: RHEL 9
- Web server: Apache httpd 2.4.53-7.el9
- Database version: mariadb-server 10.5.16-2.el9_0
- PHP version: 8.1.8-1module+el91.0+15877+c3862210
- phpMyAdmin version: 5.2.0-1.el9.remi

### Client configuration

- Browser: Firefox 108.0.1
- Operating system: Windows 10 Enterprise 20H2

### Additional context

This happens with multiple databases, each of  witch have a stored procedure that modifies SQL data.

iifawzi:
Fix #406
Signed-off-by: Fawzi Abdulfattah <iifawzie@gmail.com>


src/Statements/CreateStatement.php
```diff
@@ -240,22 +240,10 @@ class CreateStatement extends Statement
...
-            'var',
-        ],
-        'MODIFIES' => [
-            4,
-            'expr',
-        ],
+        'CONTAINS' => 4,
+        'NO' => 4,
+        'READS' => 4,
+        'MODIFIES' => 4,
```
iifawzi, line 246: IMHO, these keywords are not representing variables,  we have no changing values here. 
it's always `NO SQL`, `CONTAINS SQL`, `REQDS SQL DATA` or `MODIFIES SQL DATA` 

now, from the parser point of view, I think we might need to re-think the way we're using the contexts, do we want to consider `NO SQL` as a single keyword/token? or a composition of two keywords/tokens ( NO and SQL), what I've gone with in this PR is considering them as a composition of two separate keywords, so we can always keep the contexts adhering to what the docs say https://dev.mysql.com/doc/refman/8.0/en/keywords.html, you can notice that It includes two separate keywords, and I think that's fine. I just needed to mention that, because we seemed to have mixed between using both ways, sometimes we add composed keywords as a single keyword in the contexts, and sometimes we add them separately. 

I mentioned that this PR can also be fixed by adding the keywords as single keywords in the contexts ( `NO SQL`, `READS SQL DATA`, MODIFIES SQL DATA` ) and this's way is actually been used since the early days of the project, but I didn't find it more effective than having the keywords defined separately. 

what do you think about this generally @williamdes?

iifawzi: https://github.com/phpmyadmin/sql-parser/blob/e6320e98e3294fa32e35d54c7c0cb0aa5483e8bc/src/Contexts/ContextMariaDb100000.php#L152-L172

Here, for example, we're considering these keywords as single keywords although they're just a composition of keywords as per the docs. I think this was used for simplifying things, but maybe for the simple stuff as what we have here, we can try to keep it adhering to the docs, right? 

I think we would be so lucky if we got  @devenbansod thoughts regarding this

williamdes: > IMHO, these keywords are not representing variables,  we have no changing values here. 
it's always `NO SQL`, `CONTAINS SQL`, `READS SQL DATA` or `MODIFIES SQL DATA` 

I also think they are a `keyword/token` as on the docs https://dev.mysql.com/doc/refman/8.0/en/create-procedure.html

```text
characteristic: {
    COMMENT 'string'
  | LANGUAGE SQL
  | [NOT] DETERMINISTIC
  | { CONTAINS SQL | NO SQL | READS SQL DATA | MODIFIES SQL DATA }
  | SQL SECURITY { DEFINER | INVOKER }
}
```

iifawzi: I think having them listed this way doesn't really mean they're single keywords, we can't find them listed on the keywords/tokens here https://dev.mysql.com/doc/refman/8.0/en/keywords.html, that's why I preferred to go with considering them multiple keywords. 

from the parser side, we can go either way, but what do you think is more accurate or effective for the users? having  `NO SQL` as a single keyword? or a composition of `NO`, and `SQL`? I think as long as we can separate the tokens, that would be better, wdyt?

This might not be that critical here, but to be sure we're following the same way in the future PRs

williamdes: I think if they are in one token on the documentation and real world uses it should remain the same. If the auto completion suggests NO the user will not understand and create invalid queries
Even for people using the lib it would easier to check that one token exists rather that verifying the the combination is all there
What do you think?

williamdes: I think if they are in one token on the documentation and real world uses it should remain the same. If the auto completion suggests NO the user will not understand and create invalid queries
Even for people using the lib it would easier to check that one token exists rather that verifying the the combination is all there
What do you think?

iifawzi: I will work on making the changes later today, to follow that

williamdes (approved):

williamdes:
I think it is cleaner this way, it required no changes except contexts

PR #420, opened 2023-02-01
Provide new unit test + fix with ALTER EVENT RENAME TO using expression

niconoe-:


tests/data/parser/parseAlterEvent9.out
```diff
@@ -0,0 +1,1213 @@
...
+                            "options": {
+                                "6": "RENAME",
+                                "7": {
+                                    "name": "TO",
+                                    "equals": false,
+                                    "expr": "new_db",
+                                    "value": "new_db"
+                                }
+                            }
+                        },
```
niconoe-, line 874: As you can see, this is wrong.

williamdes: That's true !

williamdes (approved):

williamdes:
Thank you !

PR #421, opened 2023-02-08
Fix #418 - ALTER EVENT statement with DEFINER=user modifier fails to be parsed

### LINKED ISSUE #418: `ALTER EVENT` statement with `DEFINER=user` modifier fails to be parsed
Looking at the MySQL Documentation, it is possible to run such query:
```sql
ALTER DEFINER = user EVENT myEvent ENABLE;
```

Nevertheless, when you try to execute it on the linter, it fails to be parsed:[^1]
```
> php bin/lint-query --query "ALTER DEFINER = user EVENT myEvent ENABLE;"
#1: Missing comma before the beginning of a new operation ALTER. (near "ENABLE" at position 37)
#2: Operation ALTER not recognized. (near "ENABLE" at position 37)
```
My suggestion would be to add the following element into `AlterStatement::$OPTIONS`: 
```php
        // `DEFINER` is also used for `ALTER EVENT`
        'DEFINER' => [
            2,
            'expr=',
        ],
```

But I don't know the impact yet.

As far as I know, [support for the DEFINER clause was added in MySQL 5.1.17 (Section 13.1.2)](https://downloads.mysql.com/docs/refman-5.1-en.a4.pdf), so I don't know how to handle this correctly regarding the fact the sql-parser should stay compatible with MySQL 5.0.

Anyway, I'll try to provide a PR soon about it.

[^1]: messages are translated from French to English manually.

niconoe-:
Fix #418

I also improved the build process of Alter statements by adding a `trim` on it. Adding unit tests about the `ALTER DEFINER=user EVENT` builder gave me the unexpected results that an unwanted space can left at the end of the query, and already existing unit tests about the build of any Alter statement seemed to add that space in the expected value for no other reason that a lack of trimming here.

To see how the `DEFINER=user` is now correctly handle, you can compare the `.out` files from the 2 commits of the PR as the 1st one is the current status (unknown tokens and parsing errors are detected), while the 2nd one doesn't have any unknown token or parsing error.

### DIFF
--- a/src/Statements/AlterStatement.php
+++ b/src/Statements/AlterStatement.php
@@ -13,6 +13,7 @@ use PhpMyAdmin\SqlParser\Token;
 use PhpMyAdmin\SqlParser\TokensList;
 
 use function implode;
+use function trim;
 
 /**
  * `ALTER` statement.
@@ -44,10 +45,10 @@ class AlterStatement extends Statement
         'OFFLINE' => 1,
         'IGNORE' => 2,
         // `DEFINER` is also used for `ALTER EVENT`
-        //'DEFINER' => [
-        //    2,
-        //    'expr=',
-        //],
+        'DEFINER' => [
+            2,
+            'expr=',
+        ],
         'DATABASE' => 3,
         'EVENT' => 3,
         'FUNCTION' => 3,
@@ -143,8 +144,10 @@ class AlterStatement extends Statement
             $tmp[] = $altered::build($altered);
         }
 
-        return 'ALTER ' . OptionsArray::build($this->options)
+        return trim(
+            'ALTER ' . OptionsArray::build($this->options)
             . ' ' . Expression::build($this->table)
-            . ' ' . implode(', ', $tmp);
+            . ' ' . implode(', ', $tmp)
+        );
     }
 }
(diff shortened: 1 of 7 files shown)


williamdes (approved):
Maybe a more complex definer would also be a good test
With @ or w/without backquotes ?

niconoe-:
> Maybe a more complex definer would also be a good test With @ or w/without backquotes ?

I'm on it, but I discovered something I wasn't expected :laughing: . I named my file `tests/data/parser/parseAlterEventWithComplexDefiners.in`, and guess what? that name contains "lex" string so the `.out` file generated is via the **lexer** and not the **parser** !!

I'll rename my file and start again, but I really do think this should be fixed on the `Tools/TestGenerator.php` file.

niconoe-:
> Maybe a more complex definer would also be a good test With @ or w/without backquotes ?

That's done. You can compare the commit [0ccef14](https://github.com/phpmyadmin/sql-parser/pull/421/commits/0ccef14b6e389695fcc77fbea57e1c0933f451eb) for the `.out` file preparation **without the DEFINER=… management**, with the commit [e9cf491](https://github.com/phpmyadmin/sql-parser/pull/421/commits/e9cf49163ba0b3970330f656c4bfc8ab3d114643) for the updated `.out` file **with the DEFINER=… management**.

williamdes:
> file generated is via the lexer and not the parser !!

😆 😱 
We need to fix this detection one day

williamdes (approved):
Awesome !

iifawzi (approved):

PR #422, opened 2023-02-17
Incorrect order of operations to parse table/db called ``

kamil-tekiela:
A very minor error when the table name is ` `` `

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -292,8 +292,8 @@ class Token
 
                 if (isset($str[0]) && (($str[0] === '`') || ($str[0] === '"') || ($str[0] === '\''))) {
                     $quote = $str[0];
-                    $str = str_replace($quote . $quote, $quote, $str);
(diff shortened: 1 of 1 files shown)


williamdes:
I think you you also add a new test for this, check out the recent PRs for examples
I can also do it if you don't have time

kamil-tekiela:
@williamdes If you have time to do it I would be obliged. I don't have much familiarity with the parser yet.

MauricioFauth (approved):

williamdes:
> @williamdes If you have time to do it I would be obliged. I don't have much familiarity with the parser yet.

I added a test in 19ce4bdda01a4aafa46a40a2b103fecdc6862f6e
And updated the data (after the fix) in the merge commit: f022373ccc22c417c50f3349046be5df8d98a41c

williamdes (commented):
💯

PR #428, opened 2023-03-08
Fix TokensList::getPrevious which was not able to reach very first token

Tithugues:
This change aims to fix a bug in the TokensList::getPrivious method which was not able to reach the very first token. As you can see in the test case, the method returned `null` before returning the `SELECT` token.

If it was not a bug, but an expected behavior, please let me know.

Maybe such a bugfix should be backported to 5.7 as well, but I can't find a maintenance branch for this version.

Thank you.

### DIFF
--- a/src/TokensList.php
+++ b/src/TokensList.php
@@ -119,7 +119,7 @@ class TokensList implements ArrayAccess
      */
     public function getPrevious(): ?Token
     {
-        for (; $this->idx > 0; --$this->idx) {
+        for (; $this->idx >= 0; --$this->idx) {
             if (
                 ($this->tokens[$this->idx]->type !== Token::TYPE_WHITESPACE)
                 && ($this->tokens[$this->idx]->type !== Token::TYPE_COMMENT)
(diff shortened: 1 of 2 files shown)


williamdes (approved):
Do you have a SQL statement that could create a regression test for us ?

> Maybe such a bugfix should be backported to 5.7 as well, but I can't find a maintenance branch for this version.

That's okay, 5.8 will replace 5.7 without any breaking changes. Like previous versions :)

Tithugues:
> Do you have a SQL statement that could create a regression test for us ?

I think that there already is one. The one used in the test class starts by `SELECT`, but as we can see in the test itself, when we are on index 2 (the `*` of the `SELECT`), then getting previous token again lead to `null`, whereas it should have lead to the `SELECT` token.

Is it clear enough? Does it make sense to you?

> That's okay, 5.8 will replace 5.7 without any breaking changes. Like previous versions :)

Thanks! I hope it will be released soon then. :pray:

PR #429, opened 2023-03-08
Add new method TokensList::getPreviousOfType + manage list of types in TokensList::getPreviousOfType and TokensList::getNextOfType

Tithugues:
These changes aims to:
* support new request for a previous token based on a set of types
* support request for a next token based on a set of types, not only one

Changes were done in a backward compatible way.

I intentionally didn't change all the `getNextOfType*` methods, as other methods have an additional filter which does not always make sense to combine with an array of types.

Additionally, this PR relies on changes done in PR https://github.com/phpmyadmin/sql-parser/pull/428.

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -109,6 +109,23 @@ class Token
      */
     public const TYPE_LABEL = 10;
 
+    /**
+     *  All tokens types
+     */
+    public const TYPE_ALL = [
+        Token::TYPE_NONE,
+        Token::TYPE_KEYWORD,
+        Token::TYPE_OPERATOR,
+        Token::TYPE_WHITESPACE,
+        Token::TYPE_COMMENT,
+        Token::TYPE_BOOL,
+        Token::TYPE_NUMBER,
+        Token::TYPE_STRING,
+        Token::TYPE_SYMBOL,
+        Token::TYPE_DELIMITER,
+        Token::TYPE_LABEL,
+    ];
+
     // Flags that describe the tokens in more detail.
     // All keywords must have flag 1 so `Context::isKeyword` method doesn't
     // require strict comparison.
--- a/src/TokensList.php
+++ b/src/TokensList.php
@@ -132,16 +132,40 @@ class TokensList implements ArrayAccess
     }
 
     /**
+     * Gets the previous token.
+     *
+     * @param int|int[] $type the type
+     *
+     * @return Token|null
+     */
+    public function getPreviousOfType($type)
+    {
+        if (!is_array($type)) {
+            $type = [$type];
+        }
+        for (; $this->idx >= 0; --$this->idx) {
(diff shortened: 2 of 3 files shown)


williamdes (approved):
Hi @Tithugues 
Are this changes used in some other PR on in of your implementations ?

niconoe-:
I'm working with @Tithugues IRL and yes, we are expecting to use this (especially the `getPreviousOfType`).

Without entering into business details, we are working on SQL files that can define metadata on SQL statements, basically in the same state of art than PHP Attributes. So, from a statement, we need to parse the bound metadata written in a bash comment (similar to PHP Attribute, and complient with MySQL/MariaDB as such metadata can't be understood by such DBMS) which is written **before** the statement declaration, so we need to go back. As `getPrevious` is ignoring comments and there's no other proper way (`--$list->idx` is not really powerfull and user-friendly) to step back, we got the feeling this feature was lacking.

But not only for our needs, we both are confident that such feature could really help into parsing some complex options or statements into sql-parser in the future, or simplify some parsing methods that are juggling with `$list->idx`.

williamdes:
> I'm working with @Tithugues IRL and yes, we are expecting to use this (especially the `getPreviousOfType`).

That's awesome to some level !!!
Is @Tithugues French too ?
So cool !
We should probably meet some day in France


> Without entering into business details, we are working on SQL files that can define metadata on SQL statements, basically in the same state of art than PHP Attributes. So, from a statement, we need to parse the bound metadata written in a bash comment (similar to PHP Attribute, and complient with MySQL/MariaDB as such metadata can't be understood by such DBMS) which is written **before** the statement declaration, so we need to go back. As `getPrevious` is ignoring comments and there's no other proper way (`--$list->idx` is not really powerfull and user-friendly) to step back, we got the feeling this feature was lacking.
> 
> But not only for our needs, we both are confident that such feature could really help into parsing some complex options or statements into sql-parser in the future, or simplify some parsing methods that are juggling with `$list->idx`.

Very nice, be sure to upstream as much possible, we would love to benefit from your feedback and implementations

niconoe-:
> That's awesome to some level !!! Is @Tithugues French too ? So cool ! We should probably meet some day in France

Yes, we are 🙂 
It would be a pleasure! Do you agree to stay in touch by mail? (I think I have yours)

> Very nice, be sure to upstream as much possible, we would love to benefit from your feedback and implementations.
I'll discuss with @Tithugues tomorrow to see if we could share an example of the benefits this PR could bring.

williamdes:
> Yes, we are slightly_smiling_face
It would be a pleasure! Do you agree to stay in touch by mail? (I think I have yours)

So cool !, for sure you can mail me at williamdes [at] wdes dot fr
Let's do a meeting, I can travel a bit

PR #432, opened 2023-03-13
Fix #430 - `ALTER TABLE … RENAME COLUMN … TO …` is not understood by the parser/linter 

### LINKED ISSUE #430: `ALTER TABLE … RENAME COLUMN … TO …` not understood by the parser/linter
When trying to parse the following SQL statement, the `AlterStatement` createed regarding this SQL is missing the "TO" option and the new name of the column.

```sql
ALTER TABLE potato RENAME COLUMN `KIND` TO `VARIETY`;
```

The linter give also the error:
```
> bin/lint-query --query 'ALTER TABLE potato RENAME COLUMN `KIND` TO `VARIETY`;'
#1: Virgule manquante avant le début d’une nouvelle opération ALTER. (near "TO" at position 40)
```
(French translation for `"#1: Missing comma before the beginning of a new ALTER operation."`).

niconoe-:
Fixes #430 

I agree that this PR adds a specific management case for the RENAME COLUMN. If you find this too specific, you could try to give me hints on how to do this in a cleaner way, as I'm unsure this could be possible.

Thanks.

### DIFF
--- a/src/Components/AlterOperation.php
+++ b/src/Components/AlterOperation.php
@@ -13,6 +13,7 @@ use function array_key_exists;
 use function in_array;
 use function is_numeric;
 use function is_string;
+use function trim;
 
 /**
  * Parses an alter operation.
@@ -381,6 +382,13 @@ class AlterOperation extends Component
                     --$list->idx;
                 }
 
+                // If the operation is a RENAME COLUMN, now we have detected the field to rename, we need to parse
+                // again the options to get the new name of the column.
+                if ($ret->options->has('RENAME') && $ret->options->has('COLUMN')) {
+                    $nextOptions = OptionsArray::parse($parser, $list, $options);
+                    $ret->options->merge($nextOptions);
+                }
+
                 $state = 2;
             } elseif ($state === 2) {
                 if (is_string($token->value) || is_numeric($token->value)) {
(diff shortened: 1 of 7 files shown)


niconoe-:
Failing checks are only due to unmodified files. I don't expect to fix them on this PR.

williamdes (approved):
Looks good !
Out of curiosity, what happens to a query that renames two columns and also renames an index ?
Is it possible ?
https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

niconoe-:
Unit tests added via a provider to easily add some more.

And after some tests, it appears that there was just an extra space in the builder but after a trim, everything works fine ;)

src/Components/AlterOperation.php
```diff
@@ -485,18 +493,27 @@ public static function parse(Parser $parser, TokensList $list, array $options =
...
 
-        $ret .= TokensList::build($component->unknown);
+        $ret .= $afterFieldsOptions . TokensList::build($component->unknown);
 
         if (isset($component->partitions)) {
             $ret .= PartitionDefinition::build($component->partitions);
         }
 
-        return $ret;
+        return trim($ret);
```
williamdes, line 516: maybe add a second param to specify what we want to trim ?

niconoe-: I expect to trim any whitespace character (defaut value) as it looks like the default behavior to me (removing whitespaces **before** comma).

williamdes: But it will also trim new lines ;)
That's okay, I had a look and the parser mostly uses trim without any other param

niconoe-: > But it will also trim new lines ;) 

I know and I'm ok with this. When the builder is asked to build a statement, I don't expect something like this:
```sql
ALTER TABLE myTable RENAME COLUMN foo TO bar
, ADD foo VARCHAR(255) NOT NULL;
```
but rather something like:
```sql
ALTER TABLE myTable RENAME COLUMN foo TO bar, ADD foo VARCHAR(255) NOT NULL;
```

Thats' a style point of view but having whitespaces (including new lines) **before** a comma is usually not expected.

> That's okay, I had a look and the parser mostly uses trim without any other param

Thanks :wink:

niconoe-:
I wasn't able to hard set onto 5.8.x so I create a new PR (#433) from a branch that starts from 5.8.x and cherry-picks my (fixed-up) commits.

williamdes:
> I wasn't able to hard set onto 5.8.x so I create a new PR (#433) from a branch that starts from 5.8.x and cherry-picks my (fixed-up) commits.
> 
> You can close this PR now, if #433 is OK for you.

Well, let's keep this one ;)
What I did is reset hard to your other branch, and force push this one to it

williamdes (approved):
All green ✅

PR #437, opened 2023-03-24
run_generators.sh: use project root as base dir

Tithugues:
Whenever you run `php` in a container, it might be difficult to know which folders / files to mount in the container. Thus, the whole current directory is usually mounted. But in this case, relying on parent folders is not possible.

### DIFF
--- a/tools/run_generators.sh
+++ b/tools/run_generators.sh
@@ -6,12 +6,12 @@ export LC_ALL=C
 # Runs the test and context generators.
 #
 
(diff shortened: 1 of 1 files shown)


niconoe-:
Tested from a local PHP 8.0 (not via a container) and no matter the current working directory, it works, so it's fine to me.

Thanks !

williamdes (approved):
🎉

niconoe- (approved):

PR #438, opened 2023-04-03
Fix way end of functions, procedures and triggers' bodies is identified

Tithugues:
@williamdes reported that https://github.com/phpmyadmin/sql-parser/blob/4bfd15c40c6e7cec3592095cd405f52931af1b10/tests/Parser/ParserLongExportsTest.php#L59                                     was incorrect.
Indeed, in this file https://github.com/phpmyadmin/sql-parser/blob/4bfd15c40c6e7cec3592095cd405f52931af1b10/tests/data/parser/parsephpMyAdminExport1.in, the variables defined in the last comments where not reported (as we can see in the unit test).

This is due to the fact that when parsing functions, procedures and triggers, the sql-parser includes everything in the body of the statements.

With the current fix, this is not the case anymore. This has an impact on existing data test sets which seemed not to be valid.


niconoe- (approved):

niconoe-:
@williamdes Code coverage is yelling but at unchanged files, like it did a few weeks ago with some PR of mine. I think therefore we shouldn't consider those "failures" and this PR is OK and could be merged, wdyt?

Thanks :slightly_smiling_face:

williamdes:
> @williamdes Code coverage is yelling but at unchanged files, like it did a few weeks ago with some PR of mine. I think therefore we shouldn't consider those "failures" and this PR is OK and could be merged, wdyt?
> 
> Thanks 🙂

Hi
Sorry for not replying earlier
I reviewed the contents and was concerned that the `DELIMITER` would maybe mess up with how we use the parser to read procedures at phpMyAdmin
So I need to test that or someone could test?
Bu it's okay for me, I think I can merge this in the following days
And maybe release the version?
We could do a 5.8.0-rc1 maybe
So we could push more stuff to 5.8 before the next phpMyAdmin release?

williamdes (approved):
Thank you for all your contributions @Tithugues !

williamdes:
> I reviewed the contents and was concerned that the DELIMITER would maybe mess up with how we use the parser to read procedures at phpMyAdmin

Oh yeah I was right, phpMyAdmin needs fixing after this PR

If you view a complex procedure that has a `;` the procedure ends at the end of the line
In edit procedures/functions view.

The fix:
```diff
diff --git a/libraries/classes/Database/Routines.php b/libraries/classes/Database/Routines.php
index e0040139ff..369f675b90 100644
--- a/libraries/classes/Database/Routines.php
+++ b/libraries/classes/Database/Routines.php
@@ -584,7 +584,7 @@ public function getDataFromName($name, $type, $all = true): ?array
             return null;
         }
 
-        $parser = new Parser($definition);
+        $parser = new Parser('DELIMITER $$' . "\n" . $definition);
 
         /**
          * @var CreateStatement $stmt
```

PR #439, opened 2023-04-12
Fix #323 - Manage COALESCE PARTITION in ALTER TABLE, rather than COALESCE

### LINKED ISSUE #323: "Missing comma before start of a new alter operation. (near COALESCE)" for valid statement
The following command:
```sql
ALTER TABLE orders 
ADD COLUMN `n_deleted_at` timestamp AS (COALESCE(deleted_at, '1980-01-01')) VIRTUAL;
```
Produces the sql-parser error "Missing comma before start of a new alter operation. (near COALESCE)". The command still goes through just fine though.

This is on version phpmyadmin 4.9.5 and 4.9.5deb2.

Tithugues:
This PR is an attempt to fix the issue https://github.com/phpmyadmin/sql-parser/issues/323.

There was a conflict between the `COALESCE PARTITION` option and the `COALESCE` function. Here is the list of PARTITION options: https://dev.mysql.com/doc/refman/5.7/en/alter-table.html

I think that all the PARTITION options should be defined this way… But I wouldn't know how to configure all the options (1, 2, 3, var, expr, …).

Fixes: #323


williamdes (approved):
Thank you !

williamdes:
Can we say it fixes #323 ?

Tithugues:
> Can we say it fixes #323 ?

This is my expectation.

PR #440, opened 2023-04-13
Fix #319 - Support ALGORITHM and LOCK options in ALTER TABLE statements

### LINKED ISSUE #319: '",": Unrecognized alter operation' when altering table with ALGORITHM and LOCK clauses
HI!

Despite I'm using your sql-parser as part of [php-sqllint](https://github.com/cweiske/php-sqllint) I think the following issue origins from sql-parser.

I have the following statement:

`ALTER TABLE XXX DROP COLUMN YYY, ALGORITHM=INPLACE, LOCK=NONE;`

Which results in the following false-positive:

` Line 9, col 65 at ",": Unrecognized alter operation.`
` Line 9, col 67 at "LOCK": A new statement was found, but no delimiter between it and the previous one.`
` Line 9, col 67 at "LOCK": Unrecognized alter operation.`

best regards!

Tithugues:
Attempt to fix issue https://github.com/phpmyadmin/sql-parser/issues/319.

Fixes: #319

### DIFF
--- a/src/Components/AlterOperation.php
+++ b/src/Components/AlterOperation.php
@@ -70,6 +70,10 @@ class AlterOperation extends Component
             1,
             'var=',
         ],
+        'ALGORITHM' => [
+            1,
+            'var=',
+        ],
         'AUTO_INCREMENT' => [
             1,
             'var=',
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Awesome, thank you !

PR #441, opened 2023-04-21
Fix #330 - Add `KEY` to `CreateDefinition`

### LINKED ISSUE #330: Support "KEY" in table creation
### Describe the bug

phpMyAdmin 4.9.3: In SQL tab, when creating a table with a key, I get red error warnings from phpMyAdmin that don't seem to be correct.

### To Reproduce

1) Open phpMyAdmin

2) Create a new database named `misc`

3) Select misc and go to the SQL tab

4) Paste in this code:
```sql
CREATE TABLE autos (
auto_id INT UNSIGNED NOT NULL AUTO_INCREMENT KEY,
make VARCHAR(128),
year INTEGER,
mileage INTEGER
);
```

You will see lines 2, 3 and 4 flagged with a red X on the left.
I think this is valid code for MySQL, and when executed it does work as expected.

The source of the problem seems to be line 2, where the error says "A comma or a closing bracket was expected. (near KEY)". The following two lines have errors also but they appear to be derivative of the first error.

### Expected behavior

I expected no errors to be flagged.

### Screenshots
Screen Shot 2020-04-20 at 8 44 26 PM
phpMyAdmin SQL Tab

### Server configuration
 

- Operating system: macOS 10.14.6
- Web server: Apache/2.2.34 (Unix) mod_wsgi/3.5 Python/2.7.13 PHP/7.4.2 mod_ssl/2.2.34 OpenSSL/1.0.2o DAV/2 mod_fastcgi/mod_fastcgi-SNAP-0910052141 mod_perl/2.0.11 Perl/v5.24.0

- Database version: libmysql - mysqlnd 7.4.2

- PHP version: 7.4.2
- phpMyAdmin version: 4.9.3

### Client configuration

- Browser: Chrome Version 81.0.4044.113 (Official Build) (64-bit)

- Operating system: macOS 10.14.6

### Additional context

Note: Using MAMP 5.7; web and db servers on same machine as client.

williamdes:
Fixes: #330

### DIFF
--- a/src/Components/CreateDefinition.php
+++ b/src/Components/CreateDefinition.php
@@ -56,6 +56,7 @@ class CreateDefinition extends Component
             'var',
         ],
         'AUTO_INCREMENT' => 3,
+        'KEY' => 4,
         'PRIMARY' => 4,
         'PRIMARY KEY' => 4,
         'UNIQUE' => 4,
--- a/tests/data/parser/parseCreateTable13.out
+++ b/tests/data/parser/parseCreateTable13.out
@@ -430,8 +430,71 @@
                             "@type": "PhpMyAdmin\\SqlParser\\Components\\OptionsArray",
                             "options": {
                                 "1": "NOT NULL",
-                                "3": "AUTO_INCREMENT"
+                                "3": "AUTO_INCREMENT",
+                                "4": "KEY"
                             }
+                        }
+                    },
+                    {
+                        "@type": "PhpMyAdmin\\SqlParser\\Components\\CreateDefinition",
+                        "name": "make",
+                        "isConstraint": null,
+                        "type": {
+                            "@type": "PhpMyAdmin\\SqlParser\\Components\\DataType",
+                            "name": "VARCHAR",
+                            "parameters": [
+                                "128"
+                            ],
+                            "options": {
+                                "@type": "PhpMyAdmin\\SqlParser\\Components\\OptionsArray",
+                                "options": []
+                            }
+                        },
+                        "key": null,
+                        "references": null,
+                        "options": {
+                            "@type": "PhpMyAdmin\\SqlParser\\Components\\OptionsArray",
+                            "options": []
+                        }
+                    },
+                    {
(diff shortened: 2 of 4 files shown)


williamdes:
What do you think about this small change @iifawzi @Tithugues @niconoe- ?
Is this the right way to add it ?
`KEY` is already in contexts
And I added the test cases in 87f1dfa6101ab8d20a3a33e40375a06baea34cbb

Tithugues:
> What do you think about this small change @iifawzi @Tithugues @niconoe- ? Is this the right way to add it ? `KEY` is already in contexts And I added the test cases in [87f1dfa](https://github.com/phpmyadmin/sql-parser/commit/87f1dfa6101ab8d20a3a33e40375a06baea34cbb)

Yes, it seems to be OK. But as we add this option, I think we should try to support `INDEX` as well (see [CREATE table definition](https://dev.mysql.com/doc/refman/8.0/en/create-table.html)).

williamdes:
> > What do you think about this small change @iifawzi @Tithugues @niconoe- ? Is this the right way to add it ? `KEY` is already in contexts And I added the test cases in [87f1dfa](https://github.com/phpmyadmin/sql-parser/commit/87f1dfa6101ab8d20a3a33e40375a06baea34cbb)
> 
> Yes, it seems to be OK. But as we add this option, I think we should try to support `INDEX` as well (see [CREATE table definition](https://dev.mysql.com/doc/refman/8.0/en/create-table.html)).

Thank you !
I will make another PR, would you mind having a look to the two other ones I did ?

niconoe- (approved):

williamdes:
Hi @Tithugues 

I have tried to find out an example about `INDEX` but I think we already support that.
See: 52608cb84bffefd81f74da6ed9c332e2eef45975

Tithugues:
> Hi @Tithugues
> 
> I have tried to find out an example about `INDEX` but I think we already support that. See: [52608cb](https://github.com/phpmyadmin/sql-parser/commit/52608cb84bffefd81f74da6ed9c332e2eef45975)

Sorry, I misread the MySQL documentation! There is no `INDEX` option to support. :pray:

PR #75, opened 2016-09-10
Add correct parsing of SET CHARACTER SET, CHARSET , NAMES statements

devenbansod:
Add correct parsing of SET CHARACTER SET, CHARSET , NAMES statements

Add related testcases.

Fix #51 
Fix #74

PR #444, opened 2023-05-01
Update OptionsArray phpdoc & add tests for `SET GLOBAL`

ansavin:
Fixed type comment in order to pass linter check while fixing issue#17951 in phpmyadmin

Linter error: `Cannot fetch property on non-object $statement->options of type array<array-key, PhpMyAdmin\SqlParser\Components\OptionsArray>`
(example in https://github.com/phpmyadmin/phpmyadmin/actions/runs/4780247897/jobs/8498639908?pr=18349)

See https://github.com/phpmyadmin/phpmyadmin/pull/18349 for more info


williamdes:
Thanks @ansavin
Can you rebase your branch onto 5.8.x?
Please do not close or open a new PR

src/Statements/SetStatement.php
```diff
@@ -82,7 +82,7 @@ class SetStatement extends Statement
     /**
      * Options used in current statement.
      *
-     * @var OptionsArray[]|null
+     * @var OptionsArray|null
```
williamdes, line 85: Needs to be checked if that's a mistake or an actual existing behavior

ansavin:
> Thanks @ansavin Can you rebase your branch onto 5.8.x? Please do not close or open a new PR

No problem, I done it just recently

src/Statements/SetStatement.php
williamdes, line 85: For the record it was added in https://github.com/phpmyadmin/sql-parser/pull/75

I see no reason for it to be an array

PR #471, opened 2023-06-02
Drop USE_UTF_STRINGS constant

kamil-tekiela:
The days of PHP 5.5 are long gone. A constant by its definition should never change value. I assume it was meant to be a config option, but of course we don't expect anyone to change some obscure line of code. So let's remove this code as it serves no purpose anymore.


kamil-tekiela:
Just so I remember, this needs to be removed from here as well https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/classes/Linter.php#L36

williamdes (approved):
Thank you

PR #475, opened 2023-06-05
Fix #447 - Support TABLE statement

### LINKED ISSUE #447: Support `TABLE` statements
https://dev.mysql.com/doc/refman/8.0/en/table.html

TABLE is a DML statement introduced in MySQL 8.0.19 which returns rows and columns of the named table.

```sql
TABLE table_name [ORDER BY column_name] [LIMIT number [OFFSET number]]
```

The TABLE statement in some ways acts like [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html). Given the existence of a table named t, the following two statements produce identical output:

```sql
TABLE t;

SELECT * FROM t;
```

> You can order and limit the number of rows produced by TABLE using ORDER BY and LIMIT clauses, respectively. These function identically to the same clauses when used with SELECT (including an optional OFFSET clause with LIMIT).

Tithugues:
Fixes: #447

To provide:
- [ ] Unit tests

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -6,6 +6,7 @@ namespace PhpMyAdmin\SqlParser;
 
 use PhpMyAdmin\SqlParser\Exceptions\ParserException;
 use PhpMyAdmin\SqlParser\Statements\SelectStatement;
+use PhpMyAdmin\SqlParser\Statements\TableStatement;
 use PhpMyAdmin\SqlParser\Statements\TransactionStatement;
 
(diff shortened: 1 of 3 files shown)


williamdes:
Hello @Tithugues 
What should we do about this PR ?

Tithugues:
> Hello @Tithugues What should we do about this PR ?

Hi. Let's close it for now. I'll see in the future if I can work on it again. :pray:

PR #481, opened 2023-06-10
Fix build method param types

kamil-tekiela:
This is broken polymorphism as the type breaks contravariance. Actually, this method seems to break every rule of OOP. So this is just a small bandaid over the larger issue, which requires a little bit more work.


williamdes (approved):
The code change looks okay

MauricioFauth (approved):

PR #483, opened 2023-06-11
Refactor build method

kamil-tekiela:
This overrides the previous two PRs. 
This isn't even 50% of what I wanted to achieve but fixing this is really difficult. As I am going to be off of the internet for the next month, I will not have time to work on this.


williamdes:
I think it needs a rebase, I marked it as draft

MauricioFauth (approved):

Issue #234, opened 2019-03-14
Missing comma on ALTER TABLE with ENUM


MasterSalomon:
Hi, 

the following statement:

```sql
ALTER TABLE `mail_template` CHANGE COLUMN `mtpl_group` `mtpl_group` ENUM('ORDER') NULL DEFAULT NULL ;
```

throws an exeception: "Missing comma before start of a new alter operation.".

When renamed 'ORDER' to 'TEST' parser works fine.

niconoe-:
After a test, #485 fixes this. No matter the operation is a `MODIFY` or a `CHANGE`.

PR #485, opened 2023-06-21
Fix #478 - `ALTER TABLE … MODIFY … ENUM('<reserved_keyword>')` is being wrongly parsed

### LINKED ISSUE #478: `ALTER TABLE … MODIFY … ENUM('<reserved_keyword>')` is being wrongly parsed.
When trying to parse a statement like this:

```sql
ALTER TABLE `MY_TABLE` 
MODIFY `FOO` INT(11) NULL,
MODIFY `MY_COLUMN` ENUM('INSERT','UPDATE','DELETE','REPLACE') NULL,
MODIFY `BAR` VARCHAR(255) NULL;
```

SQL-Parser fails because it considers each value of the enum as a real keyword, instead of a simple string. Hence, next operations on the `ALTER TABLE` statements are not understood correctly, and it makes fail the parser.
This causes:
```
#n: A new statement was found, but no delimiter between it and the previous one. (near "&#039;INSERT&#039;" at position xxx)
```

Thanks :slightly_smiling_face:

niconoe-:
Fixes: #478
Fixes: #234


niconoe-:
@williamdes 
Base branch is `master`. As this is a fix, do you want me to cherry-pick those commit into a new branch from 5.8.x?

src/Components/AlterOperation.php
```diff
@@ -429,6 +429,13 @@ public static function parse(Parser $parser, TokensList $list, array $options =
                         $list->idx++; // Ignore the current token
                         $nextToken = $list->getNext();
 
+                        if ($token->type === Token::TYPE_STRING) {
+                            // To avoid string options, like in ENUMs, to be interpreted as statements when they
+                            // actually are just strings.
+                            $ret->unknown[] = $token;
+                            continue;
+                        }
```
williamdes, line 437: I see that below SET is handled, why ENUM is not handled like SET ?

niconoe-: I don't know, I'm not fluent with `SET()`. If those are the same, then I can merge that to get the same behavior.

williamdes:
> @williamdes Base branch is `master`. As this is a fix, do you want me to cherry-pick those commit into a new branch from 5.8.x?

Nevermind, I will rebase this onto 5.x and do the merging stuff
For further non breaking changes please use 5.x branches ;)

williamdes (approved):
Hmm, now it's unknown tokens
will this make lint errors or is this just unsupported options or the alter statement?
Could the drt or enum be included into an Expression object?

niconoe-:
> Hmm, now it's unknown tokens will this make lint errors or is this just unsupported options or the alter statement? Could the set or enum be included into an Expression object?
> 
> From what I understand this was already the behavior but it was just not that much explicit in the out files

Unfortunately, on the AlterStatement, there are no properties about the type of the column to alter in the current operation. As far as I know, we have :
- `options` to define the type of operation (ADD, MODIFY, DROP, …)
- `field` that contains basic field information, usually just the name
- `partitions` irrelevant here
- `unknown` for all tokens in the AlterOperation until the next operation or statement.

The expression representing the type of the column could be stored in the `field` property but that would have a huge impact, and even if I like the idea (as the parser will be closer to the actual statement), I'm not a fan of doing it in this PR, as it goes far beyond the problem this PR is trying to fix.

niconoe-:
Fixes also #234 (#478 was a kind of duplicate of #234 actually)

williamdes:
Hi @niconoe- 

I just noticed the company you work for is in your commit email, is it intended ?

I added 9838a1c8337c7bad71dc79676a3d992c7802fc78 to this PR

Everything seems okay, let's merge ?

niconoe-:
> I just noticed the company you work for is in your commit email, is it intended ?

:sweat_smile: It's not, but that's OK, there's no trouble with that.


Everything's fine to me! Thanks a lot!

williamdes:
> > I just noticed the company you work for is in your commit email, is it intended ?
> 
> 😅 It's not, but that's OK, there's no trouble with that.
> 
> Everything's fine to me! Thanks a lot!

Seems like that's since a long time on your contributions, feel free to change it on your git settings.
You can change this per repo ;)

Merging 🚀

PR #486, opened 2023-06-21
Fix #306 - ":=" was not recognized as an operator just like "="

### LINKED ISSUE #306: Syntax error in subquery when used in SET
Hello,

I'm using the linter provided in this library and found out that when we use it on this query `SET @test := (SELECT 1);` it fails when we try to build the object `\PhpMyAdmin\SqlParser\Statements\SetStatement` from `\PhpMyAdmin\SqlParser\Parser` with the message "A new statement was found, but no delimiter between it and the previous one.".

The parser fails also when we try to build the `\PhpMyAdmin\SqlParser\Statements\SelectStatement` because we don't add `;` at the end before the `)`, the error message is "Unexpected token.".

Both of the errors are thrown by `\PhpMyAdmin\SqlParser\Statement::parse()`.

I'm happy to create a PR with a fix for this but I'm not entirely sure how the parsing works, can someone explain it to me and suggest a fix for this?

niconoe-:
Fixes #306 as ":=" was not recognized as an operator just like "=" was.


niconoe-:
@williamdes This time, I started my branch from 5.8.x :wink:

niconoe-:
@williamdes, about the 2 failing tests, I have the feeling that's not on my hands: could it be the 5.8.x branch already having the issue even without my fix?

Maybe another PR is required before this one (and then, I can rebase) to fix those unit tests?

Thanks.

williamdes:
> @williamdes, about the 2 failing tests, I have the feeling that's not on my hands: could it be the 5.8.x branch already having the issue even without my fix?
> 
> Maybe another PR is required before this one (and then, I can rebase) to fix those unit tests?
> 
> Thanks.

You need to get the target branch right the first time or the CI will be run with the branch of your PR merged into the target branch
That's what created this inconsistent output

Closing and re-opening clears up the CI 🎉

niconoe-:
> > @williamdes, about the 2 failing tests, I have the feeling that's not on my hands: could it be the 5.8.x branch already having the issue even without my fix?
> > Maybe another PR is required before this one (and then, I can rebase) to fix those unit tests?
> > Thanks.
> 
> You need to get the target branch right the first time or the CI will be run with the branch of your PR merged into the target branch That's what created this inconsistent output
> 
> Closing and re-opening clears up the CI tada

I'm not sure I get it :laughing: . I started from the up-to-date 5.8.x, and the target branch is the same, so I don't understand why there's such issue.

Even with the current failing test, is this OK for you? Or do you need me to fix something?

williamdes:
> I'm not sure I get it . I started from the up-to-date 5.8.x, and the target branch is the same, so I don't understand why there's such issue.

You did open the PR with target to master, so GitHub did an automatic merge with master and ran the CI on the result.

master branch renamed properties to camelCase. That's how I understood what got wrong

> Even with the current failing test, is this OK for you? Or do you need me to fix something?

I only see green now 👼🏻

PR #487, opened 2023-06-22
Fix #283 - MariaDB window function with alias gives bad linting errors

### LINKED ISSUE #283: SQL query editor: MariaDB window function with alias gives bad linting errors
### Describe the bug

A MariaDB window function used together with an alias results in bad linting errors in the SQL query editor.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config/index.php?route=/table/sql&db=opengis&table=world_borders&server=1
2. Enter query:
```sql
SELECT *, COUNT(*) OVER(PARTITION BY `REGION`) AS 'count(REGION)' FROM `world_borders`
```
=> Linting errors:

phpmyadmin-linting-error

### Expected behavior

No error. Query is correct and executes fine.

### Server configuration

- phpMyAdmin version: latest (5.1.0)

### Client configuration

- Browser: Firefox
- Operating system: Windows 10

### Additional context

The error only shows up if there is an ```AS 'label'```.

niconoe-:
Fixes #283

### DIFF
--- a/src/Components/Expression.php
+++ b/src/Components/Expression.php
@@ -367,7 +367,8 @@ class Expression extends Component
                     && (($prev[1]->type === Token::TYPE_STRING)
                         || ($prev[1]->type === Token::TYPE_SYMBOL
                             && ! ($prev[1]->flags & Token::FLAG_SYMBOL_VARIABLE))
-                        || ($prev[1]->type === Token::TYPE_NONE))
+                        || ($prev[1]->type === Token::TYPE_NONE
+                            && $prev[1]->token !== 'OVER'))
                 ) {
                     if (! empty($ret->alias)) {
                         $parser->error('An alias was previously found.', $token);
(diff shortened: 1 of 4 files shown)


niconoe-:
The issue was that both "AS" and "OVER" are of type "NONE", so we were entering in the alias handling too early. The "OVER" was considered as the alias part, so when the "AS" comes out, the parser wrongly interpreted it as a second alias, and failed.

With this quickfix, I force the parser to disallow "OVER" as an alias-setting keyword, so that the whole windowing process in included in the expression, and "AS" is correctly recognized as the first (and only) alias-setting keyword.

Commits are showing the difference of the tokens management, by comparing before and after.

@williamdes This time, I picked up the right branch when I created the PR, and the tests are passing :wink:

williamdes (approved):
Looks okay, thank you !

PR #488, opened 2023-07-30
Reset sql mode after context tests

MoonE:
Some tests failed with this seed:
`composer phpunit -- --random-order-seed=1690673874`


williamdes (approved):
This should go to QA also, thank you

PR #501, opened 2023-08-29
Escape identifiers that need it

kamil-tekiela:
This should prevent more issues with unescaped identifiers. 

Ref https://dev.mysql.com/doc/refman/8.0/en/identifiers.html

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -10,6 +10,7 @@ use function in_array;
 use function intval;
 use function is_int;
 use function is_numeric;
+use function preg_match;
 use function str_replace;
 use function str_starts_with;
 use function strlen;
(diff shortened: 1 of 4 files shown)


src/Context.php
```diff
@@ -727,4 +732,9 @@ public static function hasMode(int|null $flag = null): bool
 
         return (self::$mode & $flag) === $flag;
     }
+
+    private static function doesIdentifierRequireQuoting(string $identifier): bool
+    {
+        return preg_match('/^[$]|^\d+$|[^0-9a-zA-Z$_\x80-\xffff]/', $identifier) === 1;
```
williamdes, line 738: what are we really searching for?

williamdes: I think it's worth having some unit tests to cover the regex cases

regex101 can help with building regex unit tests and then paste them into PHP code

kamil-tekiela: I have already added 4. Is there anything in particular I should build a unit test for?

MauricioFauth (approved):

src/Context.php
williamdes: Special characters would be nice, I will try to remember and push a commit

williamdes:
This should have been in QA?

kamil-tekiela:
> This should have been in QA?

Not sure. We can move it to QA if you think it's wise.

PR #504, opened 2023-08-30
Refactor TokenList

kamil-tekiela:
Tokens class was not used anywhere in PMA project. It's very likely broken. It has serious feature envy on TokensList. The two methods had very unclear purpose and behaviour.

The `build()` method in TokensList is now split into static and non-static variants. The reason for this is that CreateStatement is ready yet to use TokensList and it has to use an array instead. To make it less confusing, the new method is called `buildFromArray` which better reflects the purpose of the static method. This requires a small change in the main PMA repo. The `build()` method never actually accepted strings. It seems it was done only out of laziness in unit tests. 

The `$length` parameter isn't used anywhere, so I see no reason to keep it around. 

Edit: I had to revert change to AlterOperation because the property is used publicly in JsonSerializer.


williamdes (approved):
Looks good to me, thank you !

MauricioFauth (approved):

PR #505, opened 2023-09-06
Make constants const

kamil-tekiela:


williamdes (approved):
consts are still overridable, right ? (by inheritance)

kamil-tekiela:
> consts are still overridable, right ? (by inheritance)

Yes, I didn't add `final` flag to any of them. But overriding constants don't play well with `phpcs` so I skip the remaining overridden properties which should be overridden constants now.

williamdes (approved):

MauricioFauth (approved):

MauricioFauth (approved):

PR #506, opened 2023-09-07
Optimize offsetGet

kamil-tekiela:
We don't actually need to call getCharLength() if we inline the operation and use mb_substr with substr. This trick fetches 4 bytes from the current byte offset because UTF-8 characters are at most 4 bytes. mb_substr will then read the first Unicode character from the selected 4 bytes.

The previous technique was actually brilliant. The ASCII map optimization is the fastest way to calculate it, but it also is quite long. This technique would be perfect if PHP had a way to avoid `ord()` call altogether. I benchmarked 3 different techniques I could come up with. The ASCII map is the fastest. But actually, the worst thing for performance is a function call. So inlining the long implementation wouldn't be wise. That's why I decided to use the `mb_substr()` technique. 

Real-life benchmarks:
Before: 
image
After:
image

As you can see this is a microoptimization, so I think we shouldn't be so worried about which one is the fastest. The difference in my benchmark is only 4ms with 10k calls. 

What do you think? Should I add a code comment explaining how this works?

### DIFF
--- a/src/UtfString.php
+++ b/src/UtfString.php
@@ -102,6 +102,12 @@ class UtfString implements ArrayAccess, Stringable
      */
     public function offsetGet($offset): string|null
     {
+        // This function moves the internal byte and character pointer to the requested offset.
+        // This function is part of hot code so the aim is to do the following
+        // operations as efficiently as possible.
+        // UTF-8 character encoding is a variable length encoding that encodes Unicode
+        // characters in 1-4 bytes. Thus we fetch 4 bytes from the current offset and then use mb_substr
+        // to get the first UTF-8 character in it. We then use strlen to get the character's size in bytes.
         if (($offset < 0) || ($offset >= $this->charLen)) {
             return null;
         }
@@ -117,6 +123,8 @@ class UtfString implements ArrayAccess, Stringable
         } elseif ($delta < 0) {
             // Rewinding.
             while ($delta++ < 0) {
+                // We rewind byte by byte and only count characters that are not continuation bytes,
+                // i.e. ASCII characters and first octets of multibyte characters
                 do {
                     $byte = ord($this->str[--$this->byteIdx]);
                 } while (($byte >= 128) && ($byte < 192));
@@ -124,7 +132,8 @@ class UtfString implements ArrayAccess, Stringable
                 --$this->charIdx;
             }
         }
-
+        
+        // Fetch the first Unicode character within the next 4 bytes in the string.
         return mb_substr(substr($this->str, $this->byteIdx, 4), 0, 1);
     }
 
--- a/src/UtfString.php
+++ b/src/UtfString.php
@@ -116,10 +116,8 @@ class UtfString implements ArrayAccess, Stringable
 
         if ($delta > 0) {
             // Fast forwarding.
-            while ($delta-- > 0) {
-                $this->byteIdx += strlen(mb_substr(substr($this->str, $this->byteIdx, 4), 0, 1));
-                ++$this->charIdx;
-            }
+            $this->byteIdx += strlen(mb_substr(substr($this->str, $this->byteIdx, 4 * $delta), 0, $delta));
+            $this->charIdx += $delta;
         } elseif ($delta < 0) {
             // Rewinding.
             while ($delta++ < 0) {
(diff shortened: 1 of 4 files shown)


MauricioFauth:
> Should I add a code comment explaining how this works?

Definitely.

MauricioFauth (approved):

tests/benchmarks/UtfStringBench.php
```diff
@@ -30,35 +29,6 @@ public function benchBuildUtfString(): void
...
 
-    /**
-     * @BeforeMethods("setUp")
-     * @Iterations(2)
-     * @Revs(2)
-     * @OutputTimeUnit("microseconds")
-     * @Assert("mode(variant.time.avg) < 800 microseconds +/- 20%")
-     * @Assert("mode(variant.time.avg) > 100 microseconds +/- 10%")
-     */
-    public function benchGetCharLength(): void
```
williamdes, line 41: Can you benchmark the function and set a baseline please ?

williamdes:
> What do you think? Should I add a code comment explaining how this works?

Could we keep the previous optimisation that should normally be the fastest one and keep yours to avoid calling ord ?

src/UtfString.php
```diff
@@ -220,13 +116,13 @@ public function offsetGet($offset): string|null
...
-            }
+            $this->byteIdx += strlen(mb_substr(substr($this->str, $this->byteIdx, 4 * $delta), 0, $delta));
+            $this->charIdx += $delta;
         } elseif ($delta < 0) {
             // Rewinding.
             while ($delta++ < 0) {
+                // We rewind byte by byte and only count characters that are not continuation bytes,
+                // i.e. ASCII characters and first octets of multibyte characters
                 do {
                     $byte = ord($this->str[--$this->byteIdx]);
```
williamdes, line 127: This one still uses ord()
Could it be changed, I am not sure it's a very used case ?

tests/benchmarks/UtfStringBench.php
kamil-tekiela: Should it not remain the same? How do I run this benchmark?

williamdes:
> No, because it's either this or that. To use the ASCII map optimization one **must** use `ord()`. My solution is only slightly faster now because I replace two function calls `ord()` and `getCharLength()` with another two `substr()` and `mb_substr()` while avoiding any `if`s. If there is a way to do this using only one function call or less then that would be the best solution.
> 
> But maybe I misunderstood what you mean. What exactly are you proposing?

Indeed, I misunderstood the new solution 
But it feels like it requires much more functions, is there no way to avoid using some of them ?
Not sure if's me over optimizing the solution ^^

> Do you see a way to merge these two solutions together?

Not really because of the use of delta and other things that seems to advance the string much more quickly


Maybe phpbench could show how this is quick ?

tests/benchmarks/UtfStringBench.php
williamdes: Well, since it's a more complete function and the function was removed maybe everything should be updated then
But having a benchmark data could help

src/UtfString.php
kamil-tekiela: It could but I doubt it's worth it. Reading a substring back is more complicated than forward because of boundary checks. This is actually pretty clean and I think it can stay as it is.

kamil-tekiela:
> But it feels like it requires much more functions, is there no way to avoid using some of them ?

Yeah it does. I wish there was a way to avoid it but I can't think of any. Good thing that `strlen` isn't a function call but a dedicated opcode.

src/UtfString.php
williamdes: Okay, let's keep it as is.
Could my function be restored and used in this place ?

kamil-tekiela: No, because that function looked at the starting octet, but when rewiring we don't know where the starting octet is.

williamdes: Okay

tests/benchmarks/UtfStringBench.php
kamil-tekiela: I added a new commit. Is this what you had in mind? These are the results I get:

```
    benchBuildUtfString.....................I19 ✔ Mo6.052ms (±3.59%)
    benchUtfStringRandomAccessWithUnicode...I19 ✔ Mo63.459μs (±10.50%)
```

williamdes (approved):
Looks great !

tests/benchmarks/UtfStringBench.php
williamdes: Yes, that's good to have checked at all times.
For the results I do not know, if the benchmark was added before we could compare but that's totally okay never mind :)

MauricioFauth (approved):

PR #499, opened 2023-08-15
Add support for set = ? for parser

BackEndTea:
Fixes: #492

PR #515, opened 2023-09-26
Fix using ? as a parameter

### LINKED ISSUE #202: doesn't support sql parameter in set clause
When use parser to parse the sql like this:
```sql
  update tbl_customer t set t.`description`=:yp where t.id=1;
```
the parser failed.it seems the parser cann't handle sql parameter in the SET sub-clause.

### LINKED ISSUE #412: Support for parameterized select clauses
When using placeholders in a select clause for a prepared statement, the parser stops at the first placeholder.  For example, parsing "SELECT 1, 2, ?, 3" will include the expressions 1 and 2 in the SelectStatement object, but not the placeholder or 3.
```
PhpMyAdmin\SqlParser\Statements\SelectStatement Object
(
    [expr] => Array
        (
            [0] => PhpMyAdmin\SqlParser\Components\Expression Object
                (
                    [database] => 
                    [table] => 
                    [column] => 
                    [expr] => 1
                    [alias] => 
                    [function] => 
                    [subquery] => 
                )

            [1] => PhpMyAdmin\SqlParser\Components\Expression Object
                (
                    [database] => 
                    [table] => 
                    [column] => 
                    [expr] => 2
                    [alias] => 
                    [function] => 
                    [subquery] => 
                )

        )

    [from] => Array
        (
        )

    [index_hints] => 
    [partition] => 
    [where] => 
    [group] => 
    [having] => 
    [order] => 
    [limit] => 
    [procedure] => 
    [into] => 
    [join] => 
    [union] => Array
        (
        )

    [end_options] => 
    [options] => PhpMyAdmin\SqlParser\Components\OptionsArray Object
        (
            [options] => Array
                (
                )

        )

    [first] => 0
    [last] => 11
)
```

### LINKED ISSUE #492: Missing support for SET = ?
The following query results in an error with the message of `Missing expression.` when its parsed. 
 
```sql
-- gives error
UPDATE  order
SET     price = ?
WHERE   orderid = ?
```

```sql
-- does not error
UPDATE  order
SET     price = 3
WHERE   orderid = ?
```

MauricioFauth:
- Fixes #202
- Fixes #412
- Fixes #492
- Closes #499
- Related to #172


tests/data/misc/parseParameter2.out
```diff
@@ -0,0 +1,366 @@
...
+                67,
+                0
+            ],
+            [
+                "Variable name was expected.",
+                "?",
+                70,
+                0
+            ]
+        ],
```
williamdes, line 363: should this still exist ?

MauricioFauth: You right. That's another issue. Thanks!

tests/data/bugs/gh492.out
```diff
@@ -0,0 +1,260 @@
...
+                0
+            ],
+            [
+                "Unrecognized keyword.",
+                {
+                    "@type": "@4"
+                },
+                0
+            ]
+        ]
```
williamdes, line 258: I think this is the only remaining bit to fix

williamdes (approved):
Awesome !
Will it be easy to detect the ? or := for our (broken) feature "Bind parameters" in the query box ?

tests/data/bugs/gh492.out
MauricioFauth: This should not be removed because the statement is actually wrong. The table name should be quoted.

williamdes: oh okay, that's why

MauricioFauth:
> Will it be easy to detect the ? or := for our (broken) feature "Bind parameters" in the query box ?

Is there an issue for that?

williamdes:
> > Will it be easy to detect the ? or := for our (broken) feature "Bind parameters" in the query box ?
> 
> Is there an issue for that?

No, I did not report it until now since the parser has unhappy

PR #522, opened 2024-01-13
Add type aliases

kamil-tekiela:
@williamdes  I don't know if I did this right or how to create tests for this. Can you please check?


williamdes (approved):
Technically I think this should be right
But needs to be confirmed that what you added is supported by the lowest MySQL and MariaDB versions here

williamdes:
This should go to the QA branch, I can rebase it

MauricioFauth:
> This should go to the QA branch, I can rebase it

@williamdes I'm gonna merge it to 5.9.x.

PR #526, opened 2024-01-15
Add native parameter types to Token constructor 

MauricioFauth:
Ensures that `token` are always a string and that `flags` are always an integer.

Using an empty string for the last delimiter has the same effect as using `null`.

- Related to #521


MauricioFauth:
@kamil-tekiela What do you think?

kamil-tekiela:
On line 256 in Lexer you still assign `null` to `$delimiter` property which is then used on line 1013. Isn't this a problem still?

src/Lexer.php
```diff
@@ -360,7 +360,7 @@ private function solveAmbiguityOnFunctionKeywords(): void
                 && ($next->type !== TokenType::Operator
                     || ! in_array($next->value, self::OPERATOR_NAME_INDICATORS, true)
                 )
-                && ($next->value !== null)
+                && ($next->value !== '')
```
kamil-tekiela, line 363: I hope that this ok, but I feel like the phpdoc should be updated.

williamdes (approved):

src/Lexer.php
MauricioFauth: Token::$value type should definitely be updated. That value is only created by Lexer line 285.
https://github.com/phpmyadmin/sql-parser/blob/94f2db2fd7dded6ba251bf15e4645b5d19b10a35/src/Lexer.php#L284-L285

MauricioFauth:
> On line 256 in Lexer you still assign `null` to `$delimiter` property which is then used on line 1013. Isn't this a problem still?

Lexer::$delimiter should only be string. But this is not an issue because it gets coerced to string right after.

PR #528, opened 2024-01-15
Add native property types in Lexer

kamil-tekiela:
I am still not sure what to do about this `null` delimiter, so at the moment I decided to just add it to the baselines.

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -239,7 +239,7 @@ class Lexer
                 $pos = $this->last + 1;
 
                 // Parsing the delimiter.
-                $this->delimiter = null;
+                $this->delimiter = '';
                 $delimiterLen = 0;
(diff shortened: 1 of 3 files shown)


src/Lexer.php
```diff
@@ -66,60 +66,46 @@ class Lexer
...
+    public static string $defaultDelimiter = ';';
 
     /**
      * Statements delimiter.
      * This may change during lexing.
-     *
-     * @var string
      */
-    public $delimiter;
+    public string|null $delimiter;
```
MauricioFauth, line 99:
```suggestion
    public string $delimiter;
```

MauricioFauth (changes requested):

MauricioFauth (approved):

PR #535, opened 2024-01-16
Components refactoring

kamil-tekiela:
My idea here is to tidy this all up by splitting the `static` functionality from the value objects. But the ultimate goal is to fix the instantiation of all Component objects, which this PR doesn't really help with. I consider this to be a good change nonetheless.


MauricioFauth (approved):
I agree with these changes.

kamil-tekiela:
> I agree with these changes.

Thanks. I rebased it. I am wondering if I should move all `parse` methods into their own classes in the `Parsers` namespace.

MauricioFauth:
> I am wondering if I should move all `parse` methods into their own classes in the `Parsers` namespace.

This is a good idea.

Issue #357, opened 2021-09-28
Condition could have more detailed fields


afilina:
Currently, when I parse a query, I get WHERE as `Condition` instances. These have an `expr` field, which is just a string and not very useful when analyzing a query. It would be really nice if we could be able to get more detail from that, like the operator, operands, etc.

Examples:

```php
// c.id = 2
[
    leftOperand: ColumnReference [ column: 'id', tableAlias: 'c' ] ,
    operator: Equals []
    rightOperand: Constant [ value: 2 ]
]
```

```php
// customer.address_id IS NOT NULL
[
    leftOperand: ColumnReference [ column: 'address_id', tableName: 'customer' ] ,
    operator: IsNotNull
    rightOperand: null
]
```

mauriciofauth:
Thank you. We will work on this.

PR #549, opened 2024-01-26
Add basic details for Condition

MauricioFauth:
Adds `leftOperand`, `operator` and `rightOperand` properties to the `Condition` component.

It only splits comparison operators for now.

- Related to #357

### DIFF
--- a/src/Parsers/Conditions.php
+++ b/src/Parsers/Conditions.php
@@ -57,6 +57,8 @@ final class Conditions implements Parseable
         'XOR',
     ];
 
+    private const COMPARISON_OPERATORS = ['=', '>=', '>', '<=', '<', '<>', '!='];
+
     /**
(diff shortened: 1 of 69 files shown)


src/Parsers/Conditions.php
```diff
@@ -57,6 +57,8 @@ final class Conditions implements Parseable
         'XOR',
     ];
 
+    private const COMPARISON_OPERATORS = ['=', '>=', '>', '<=', '<', '<>', '!='];
```
kamil-tekiela, line 60: Why is [`<=>`](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to) not here?

MauricioFauth: This operator has a syntax a little bit different, so I didn't add it yet.

PR #553, opened 2024-03-01
Add support for `ALTER FUNCTION` and `ALTER PROCEDURE` statements

niconoe-:
Add support for `ALTER FUNCTION` and `ALTER PROCEDURE` statements.

The ABNF for those statements are exactly the same since MySQL 5.0 to latest MariaDb version:
* Doc for MySQL 5.0: https://downloads.mysql.com/docs/refman-5.0-en.a4.pdf#page=1083
* Doc for MySQL 8.0: https://dev.mysql.com/doc/refman/8.0/en/alter-function.html
* Doc for MariaDb: https://mariadb.com/kb/en/alter-function/

2 commits are done: the first with everything but the options definitions in case of `ALTER PROCEDURE` or `ALTER FUNCTION`, causing all `*.out` data files giving the error "Unrecognized alter operation." we are trying to solve. The second commit applies the options definitions, so you can see that the related `*.out` files are updated to no longer have the parser error, but the full options built instead.

@williamdes It's been a long time :smile: . I'll not be back for so long, but I never come back with empty hands :laughing: .


williamdes (approved):
Thank you, this all looks good to me
@kamil-tekiela what do you think ?

kamil-tekiela (approved):
It looks fine.

niconoe-:
I'm glad you valid this 🙂 

I'm having the Benchmark on PHP 8.0 not passing, but I don't understand why. Is this a blocking point preventing this PR to be merged? If yes, how can I handle this please?

Thanks a lot!

PR #554, opened 2024-04-18
Allow parsing ALTER TABLE statement with column check constraint

MoonE:
`CHECK` is in Parser::STATEMENT_PARSERS, but also allowed in the column definition.

This doesn't enforce an order for the check constraint. It seems to be only allowed at the end of the column definition, only the column order (`FIRST` / `AFTER x`) can follow.


williamdes (approved):

tests/data/parser/parseAlterTableAddColumnWithCheck.out
```diff
@@ -0,0 +1,428 @@
...
+                            "database": null,
+                            "table": null,
+                            "column": "json",
+                            "expr": "`json`",
+                            "alias": null,
+                            "function": null,
+                            "subquery": null
+                        },
+                        "partitions": null,
+                        "unknown": [
```
williamdes, line 346: I guess we do not have proper parsing of the check expression

MauricioFauth (approved):

PR #555, opened 2024-04-28
Fix replace clause of select statement with FOR UPDATE

MoonE:
Fixes phpmyadmin/phpmyadmin#17190

Replaces _END_OPTIONS with the actual names from here:
https://github.com/phpmyadmin/sql-parser/blob/29f982a9559ac9a97b19e58b78a3fa4c41d5e43f/src/Statements/SelectStatement.php#L84-L228

### DIFF
--- a/src/Statement.php
+++ b/src/Statement.php
@@ -9,6 +9,7 @@ use Stringable;
 
 use function array_flip;
 use function array_keys;
+use function array_push;
 use function count;
 use function in_array;
 use function stripos;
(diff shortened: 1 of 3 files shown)


src/Statement.php
```diff
@@ -457,6 +458,28 @@ public function getClauses()
...
+     * @return array<string, int>
+     */
+    public function getClauseOrder(): array
+    {
+        $clauses = [];
+        foreach (array_keys($this->getClauses()) as $key) {
+            if ($key === '_END_OPTIONS') {
+                if (static::$END_OPTIONS !== []) {
+                    array_push($clauses, ...array_keys(static::$END_OPTIONS));
+                }
```
MoonE, line 474: When mergin to master (PHP >= 7.3) the if is no long needed
```suggestion
                array_push($clauses, ...array_keys(static::$END_OPTIONS));
```
https://www.php.net/manual/en/function.array-push.php

src/Statement.php
```diff
@@ -457,6 +458,28 @@ public function getClauses()
...
+     * Gets the clause order of this statement as an array
+     * with clause as key and index as value.
+     *
+     * @return array<string, int>
+     */
+    public function getClauseOrder(): array
+    {
+        $clauses = [];
+        foreach (array_keys($this->getClauses()) as $key) {
+            if ($key === '_END_OPTIONS') {
```
williamdes, line 471: When merging in master, this key changes name

williamdes (approved):

MauricioFauth (approved):

PR #557, opened 2024-05-01
Allow parsing KILL statement

### LINKED ISSUE #556: Unrecognized statement type. (near KILL)
### Describe the bug

This is a valid query, but it fails parsing it.

```SQL
KILL 2883486;
```

### To Reproduce

1. Go to https://demo.phpmyadmin.net/QA_5_2/index.php?route=/server/sql
2. Paste `KILL 2883486;`
3. See error

### Expected behavior

No error should happen.

### Screenshots

kill

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

MoonE:
Fixes #556


MoonE:
Could also use `$CLAUSES` to build it. Though it fails to build an invalid statement like `KILL ;` which may not matter.
```patch
diff --git a/src/Statements/KillStatement.php b/src/Statements/KillStatement.php
index 7040a0f3..4ef3f532 100644
--- a/src/Statements/KillStatement.php
+++ b/src/Statements/KillStatement.php
@@ -28,16 +28,22 @@ class KillStatement extends Statement
         'QUERY' => 1,
     ];
 
+    public static $CLAUSES = [
+        'KILL' => [
+            'KILL',
+            2,
+        ],
+        // Used for options.
+        '_OPTIONS' => [
+            '_OPTIONS',
+            1,
+        ],
+        'EXPRESSION' => [
+            'KILL',
+            1,
+        ],
+    ];
+
     /** @var Expression|null */
     public $processListId = null;
-
-    public function build(): string
-    {
-        $option = $this->options === null || $this->options->isEmpty()
-            ? ''
-            : ' ' . OptionsArray::build($this->options);
-        $expression = $this->processListId === null ? '' : ' ' . Expression::build($this->processListId);
-
-        return trim('KILL' . $option . $expression);
-    }
 }
```

williamdes (approved):
Looks good to me
@niconoe- or @iifawzi could you also review please ?

niconoe-:
LGTM but I don’t know the KILL command 😅

src/Statements/KillStatement.php
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Options of this statement.
+     *
+     * @var array<string, int|array<int, int|string>>
+     * @psalm-var array<string, (positive-int|array{positive-int, ('var'|'var='|'expr'|'expr=')})>
+     */
+    public static $OPTIONS = [
+        'CONNECTION' => 1,
+        'QUERY' => 1,
+    ];
```
iifawzi, line 29: It would be awesome if we could extend support of[ Maria KILL options ](https://mariadb.com/kb/en/kill/#:~:text=You%20can%20see%20which%20threads,given%20thread%20or%20query%20id.)

src/Statements/KillStatement.php
```diff
@@ -0,0 +1,43 @@
...
+use PhpMyAdmin\SqlParser\Statement;
+
+use function trim;
+
+/**
+ * `KILL` statement.
+ *
+ * KILL [CONNECTION | QUERY] processlist_id
+ */
+class KillStatement extends Statement
```
iifawzi, line 18: I do believe we need some parsing logic here, so we can accurately validate the statement

tests/Parser/KillStatementTest.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * @return array<int, array<int, string>>
+     * @psalm-return list<list<string>>
+     */
+    public function buildKillProvider(): array
+    {
+        return [
+            ['KILL (SELECT 3 + 4)'],
```
iifawzi, line 53: is it a valid use case? why would someone need to kill a query or connection that way?

tests/Parser/KillStatementTest.php
```diff
@@ -0,0 +1,59 @@
...
+            ['KILL QUERY 3'],
+            ['KILL CONNECTION 3'],
+            ['KILL'],
```
iifawzi, line 56: This should be invalid, hence why the parsing logic is so important

iifawzi (commented):
Thanks for your hard work! it looks good to me, just if we could have the parsing logic it would be awesome!

tests/Parser/KillStatementTest.php
iifawzi, line 56: I meant `KILL` is what expected to be invalid

tests/Parser/KillStatementTest.php
MoonE: The use case may be questionable, but the query is valid and working.

iifawzi: or maybe someone can utilize this by dynamically killing queries that are locking others? by selecting their IDs from the activity? if that's the case, maybe we can also parse the nested select statements? and delegate the building of these queries to their classes? ( most probably it's only SELECT statement that would be allowed ). what do you think?

src/Statements/KillStatement.php
MoonE: Oh yeah, makes sense.

tests/Parser/KillStatementTest.php
MoonE: This is parsed with an error, I was unsure whether it should be able to *build* a query that was parsed with error. It is probably not needed.

iifawzi: It's been a while since I interacted with the parser, so I might be mistaken. How this's parsed with an error? if there's no parse function I don't think the statement would be even parsed or validated

iifawzi: image

I believe we need the parsing logic so we can throw an error if it's incorrect, the same as we did with the `CREATE` statement for example

src/Statements/KillStatement.php
```diff
@@ -0,0 +1,43 @@
...
+     * @var array<string, int|array<int, int|string>>
+     * @psalm-var array<string, (positive-int|array{positive-int, ('var'|'var='|'expr'|'expr=')})>
+     */
+    public static $OPTIONS = [
+        'CONNECTION' => 1,
+        'QUERY' => 1,
+    ];
+
+    /** @var Expression|null */
+    public $processListId = null;
```
iifawzi, line 32: I believe we should instead utilize `var` options instead of using custom option

```PHP
    public static $OPTIONS = [
        'CONNECTION' => [
            1,
            'var=',
        ],
        'QUERY' => [
            1,
            'var=',
        ],
    ];
```

MauricioFauth (approved):

PR #558, opened 2024-05-09
Don't add too many spaces when replacing clauses

MoonE:
There were unnecessary spaced included in the statement when
- a clause is removed (replaced with empty string)
- the replacement string already has spaces 
- the replacement was at the start or end of the statement

This will require some tests to be updated in phpMyAdmin.
e. g.: https://github.com/phpmyadmin/phpmyadmin/blob/99c2939eda6c1c3ab9ffe4b1f903bab182d686db/test/classes/Display/ResultsTest.php#L1469

### DIFF
--- a/src/Utils/Query.php
+++ b/src/Utils/Query.php
@@ -33,8 +33,10 @@ use PhpMyAdmin\SqlParser\TokenType;
 use function array_flip;
 use function array_keys;
 use function count;
+use function ctype_space;
 use function in_array;
 use function is_string;
+use function mb_substr;
 use function trim;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Utils/Query.php
```diff
@@ -726,6 +727,27 @@ public static function getClause($statement, $list, $clause, $type = 0, $skipFir
...
+     */
+    private static function glueQueryPartsWithSpaces(array $parts): string
+    {
+        $statement = '';
+        foreach ($parts as $part) {
+            if ($part === '') {
+                continue;
+            }
+
+            if ($statement !== '' && ! preg_match('/\\s$/', $statement) && ! preg_match('/^\\s/', $part)) {
```
williamdes, line 741: I would like to avoid this function
is there another way with a simple function
maybe str ends with?

MoonE: [`ctype_space`](https://www.php.net/manual/en/function.ctype-space.php) on the character in question should work.

williamdes (approved):
Nice, this version is probably better because I see quite some crash reports due to ram eaten while trying to compile a regex

MauricioFauth:
I think this should target the `master` branch.

williamdes (approved):

MauricioFauth (approved):

PR #560, opened 2024-05-31
Fix where 0

MoonE:
If a part of a condition is `0` it was not parsed correctly


kamil-tekiela (approved):
Perfect! There is no way for this variable to be anything other than a string at this point.

MauricioFauth (approved):

PR #564, opened 2024-07-03
Fix query flags for lower-case functions

MoonE:
See https://github.com/phpmyadmin/phpmyadmin/issues/16869

`is_func` and `is_count` were only set when the function is in all upper case.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -956,6 +956,11 @@ parameters:
 			path: src/Utils/Query.php
 
 		-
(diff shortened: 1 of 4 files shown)


williamdes (approved):
@iifawzi

iifawzi (approved):

iifawzi:
it's unexpectedly faster

williamdes:
> it's unexpectedly faster? What these numbers are based on?

Oh, on what I had when I wrote the bench. Maybe we can update to quicker then

PR #565, opened 2024-07-09
Fix #556 - Implementing and refactoring the KILL statement parser

### LINKED ISSUE #556: Unrecognized statement type. (near KILL)

iifawzi:
Hey! 

This PR extends support of KILL statements. completing #557 and fixing [556](https://github.com/phpmyadmin/sql-parser/issues/556)

Fixes #556


.gitignore
```diff
@@ -15,3 +15,4 @@ phpstan.neon
 # Infection
 infection.json
 infection.log
+**/.DS_Store
```
kamil-tekiela, line 18: This doesn't look like it should be part of this pr

src/Statements/KillStatement.php
```diff
@@ -24,20 +30,153 @@ class KillStatement extends Statement
...
+     * @psalm-var Statement|int|null
+     */
+    public $identifier = null;
+
+    /**
+     * Whether MariaDB ID keyword is used or not.
+     *
+     * @psalm-var bool
+     */
+    public $IDKeywordUsed = false;
```
kamil-tekiela, line 52: Why are the property types missing? Can you please add all of them?

src/Statements/KillStatement.php
```diff
@@ -24,20 +30,153 @@ class KillStatement extends Statement
...
+                        break 2;
+                    }
+
+                    break;
+
+                case 3:
+                    if ($token->type === Token::TYPE_KEYWORD && $token->value === 'SELECT') {
+                        $subList = new TokensList(array_slice($list->tokens, $list->idx - 1));
+                        $subParser = new Parser($subList);
+                        if (count($subParser->errors)) {
```
kamil-tekiela, line 122: Please don't use count() as a Boolean value. Use strict array comparison. !== []

What exactly is the purpose of this if statement though?

src/Statements/KillStatement.php
```diff
@@ -24,20 +30,153 @@ class KillStatement extends Statement
...
+     */
+    public function build()
     {
-        $option = $this->options === null || $this->options->isEmpty()
-            ? ''
-            : ' ' . OptionsArray::build($this->options);
-        $expression = $this->processListId === null ? '' : ' ' . Expression::build($this->processListId);
+        $ret = 'KILL';
+
+        if ($this->options && count($this->options->options) > 0) {
```
kamil-tekiela, line 161: `$this->options` is not a Boolean. Please add the condition. And same as above please don't use count as a Boolean.

src/Statements/KillStatement.php
```diff
@@ -24,20 +30,153 @@ class KillStatement extends Statement
...
+            $ret .= ' ' . OptionsArray::build($this->options);
+        }
+
+        if ($this->IDKeywordUsed) {
+            $ret .= ' ID';
+        }
+
+        $builtIdentifier = (string) $this->identifier;
```
kamil-tekiela, line 169: Isn't this the same as the code in the if statement below?

kamil-tekiela (changes requested):

.gitignore
iifawzi: it isn't worth it to be in a separate PR. what do you think?

src/Statements/KillStatement.php
iifawzi: phpstan complains about the php7 support. should we add it and update the baseline?

src/Statements/KillStatement.php
iifawzi: sure, way more performant. thanks for the suggestion. 
checking the errors of subparser is needed so we can throw them. do u have better ways in mind?

src/Statements/KillStatement.php
kamil-tekiela: @MauricioFauth what is the minimum version of PHP? Is it 7.2?

src/Statements/KillStatement.php
iifawzi: Indeed, I didn't notice that `Statement` implements`__toString`

src/Statements/KillStatement.php
iifawzi: interestingly, in `5.10.x` it's 7.2 yes. but in the main, it's 8.1..

iifawzi: I might have forgotten how we do releases, shouldn't 5.10.x be the upcoming version? how it supports 7.2 and main isn't?

tests/data/parser/parseKill2.out
```diff
@@ -0,0 +1,189 @@
...
+    "parser": {
+        "@type": "PhpMyAdmin\\SqlParser\\Parser",
+        "list": {
+            "@type": "@1"
+        },
+        "statements": [
+            {
+                "@type": "PhpMyAdmin\\SqlParser\\Statements\\KillStatement",
+                "identifier": {
+                    "@type": "PhpMyAdmin\\SqlParser\\Statements\\SelectStatement",
```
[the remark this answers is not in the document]
williamdes: I am not sure what to pick
Can you share a diff or something more explicit to help choosing ?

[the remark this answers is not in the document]
iifawzi: it's either the built statement: 
```
  "identifier": "SELECT 3 + 4",
``
or the statement object ( Select Statement ) as committed in the file

**Comment by kamil-tekiela** [comment]
DS_Store are MacOS files so they are not relevant to this project. They should not be ignored at all. You can ignore them locally on your machine.

**Comment by iifawzi** [comment]
it's easier for future contributors to have it ignored by the repo, like we do with .idea files.

File: src/Statements/KillStatement.php (hunk 2 above)
Comment (MauricioFauth) [reply src/Statements/KillStatement.php:51]: The version 5.x of sql-parser are used by phpMyAdmin 5.2.x, that's why it requires PHP 7.2, the `master` branch is for sql-parser 6.x, which is used by phpMyAdmin 6.0.x.

File: tests/data/parser/parseKill2.out (hunk 6 above)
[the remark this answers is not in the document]
Comment (williamdes) [reply tests/data/parser/parseKill2.out:137]: I think Select Statement has better parsing for many possible syntaxes
And it can be ->toString() that should equal to the old data ("identifier") ?

File: src/Statements/KillStatement.php (hunk 2 above)
Comment (williamdes) [reply src/Statements/KillStatement.php:51]: yes, or the merge user can add them at merge time

**Comment by williamdes** [comment]
> DS_Store are MacOS files so they are not relevant to this project. They should not be ignored at all. You can ignore them locally on your machine.

I checked, and yes I think the actual consensus is to remove them locally. We do not have it in any repo.

Ref: https://github.com/phpmyadmin/phpmyadmin/issues/15903#issuecomment-581431150
Can you remove the commit ?

File: tests/data/parser/parseKill2.out (hunk 6 above)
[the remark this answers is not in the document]
Comment (iifawzi) [reply tests/data/parser/parseKill2.out:137]: Yes, exactly.

File: src/Parser.php [hunk 7]
```diff
@@ -211,10 +211,6 @@ class Parser extends Core
             'class' => 'PhpMyAdmin\\SqlParser\\Components\\JoinKeyword',
             'field' => 'join',
         ],
-        'KILL' => [
-            'class' => 'PhpMyAdmin\\SqlParser\\Components\\Expression',
```
MauricioFauth, line 215: I think creating a `PhpMyAdmin\SqlParser\Components\KillKeyword` class will be better than overriding `Statement::parse()`.

iifawzi: Yes, that can certainly be done, and we can rely on the clauses to have it parsed. Correct me if you meant another way.

https://github.com/phpmyadmin/sql-parser/blob/417ad4cf2eaa13d61dfd67dacf0458c9b586be5d/src/Statement.php#L263-L274

I just believe that components should be used for parts that are common across multiple query statements, which aligns with the current behavior of our components. What do you think? In my opinion, "KILL" is a statement and shouldn't be identified as a keyword.

Given this discussion, I think conventions have become mixed over the years. That's why I would really appreciate everyone's participation in this discussion to ensure consistency moving forward.

MauricioFauth: > I just believe that components should be used for parts that are common across multiple query statements

That makes sense.

MauricioFauth (approved):

src/Statements/KillStatement.php
```diff
@@ -24,20 +29,149 @@ class KillStatement extends Statement
...
+     *
+     * @var bool
+     */
+    public $idKeywordUsed = false;
 
-    public function build(): string
+    /**
+     * Whether parenthesis used around the identifier or not
+     *
+     * @psalm-var bool
```
williamdes, line 56:
```suggestion
     * @var bool
```

niconoe- (approved):
For what it worth, LGTM.

PR #568, opened 2024-08-02
Define context keywords relying on Token constants

Tithugues:
I always had issues to understand the "magic" keyword integers of the Context* files. Would it make sense to rely on Token constants to define them?

Additionally, I added a variable for the default bit needed for all keywords.


M393:
These files are autogenerated with https://github.com/phpmyadmin/sql-parser/blob/master/tools/run_generators.sh

This is the code that prints the keywords:
https://github.com/phpmyadmin/sql-parser/blob/3bc8d0dde4daaee6d074349655a248f3ee88937b/src/Tools/ContextGenerator.php#L234-L247

Tithugues:
OK, thanks. Then I’ll take a look at the generator. 😅

PR #569, opened 2024-08-06
For Contexts, rely on token constants + Add MySQL 8.4 context

niconoe-:
**EDIT: I'll probably close this one in profit of #570** 

I'm taking this from @Tithugues, enhancing the generator itself rather than doing the job "manually".

Main changes are:
* Stop using magic numbers to use const instead in both generators and generated contexts
* As generated contexts are now formatted in a way each word has its own line, simplify the builder
  * Words are still sorted by types, then by alphanumerical order, but no longer by length, as it no longer make sense
  * Some comments/phpDoc are fixed on the generated contexts
* Fix reserved keywords from MariaDB 10.3, based on the list given at [https://mariadb.com/kb/en/reserved-words/](https://mariadb.com/kb/en/reserved-words/)
* Add MySQL 8.4 context, using changes described on [https://dev.mysql.com/doc/refman/8.4/en/keywords.html](https://dev.mysql.com/doc/refman/8.4/en/keywords.html) to make the list.

Please let me know if something is wrong or is missing :wink:


kamil-tekiela (approved):
I like it. I wanted to propose something similar myself.

williamdes (approved):
I would say we want the changes for target branch 5.10.0 ?

niconoe-:
> I would say we want the changes for target branch 5.10.0 ?

I don't, know TBH. I targeted `master` just as @Tithugues did on #568 , but if you want this on another tag, I'll adapt.

williamdes:
> > I would say we want the changes for target branch 5.10.0 ?
> 
> I don't, know TBH. I targeted `master` just as @Tithugues did on #568 , but if you want this on another tag, I'll adapt.

Well, that's okay for me both ways
But the changes will not be released very soon but only when the major version will be out

niconoe-:
> > > I would say we want the changes for target branch 5.10.0 ?
> > 
> > 
> > I don't, know TBH. I targeted `master` just as @Tithugues did on #568 , but if you want this on another tag, I'll adapt.
> 
> Well, that's okay for me both ways But the changes will not be released very soon but only when the major version will be out

To me, that's not a BC Break change at all, so it shouldn't be requested for the next major.
And to be honest, beside the addition of the MySQL 8.4 context, there's nothing new too, so we might even consider this as a patch :laughing: 

What's the target branch if I want to propose it for the next 5.9 ?

williamdes:
> What's the target branch if I want to propose it for the next 5.9 ?

https://github.com/phpmyadmin/sql-parser/tree/5.9.x

That will be merged up to master

niconoe-:
Closing this one in profit of #570

PR #570, opened 2024-08-06
Feature/from 5 9 define context keywords relying on token constants

niconoe-:
Same as #569 but from 5.9.x branch.


niconoe-:
@williamdes / @kamil-tekiela 
Are you OK with these changes? I'm having issues with PHPStan about files I didn't change, and with Psalm on Generated contextes but this will be fixed on future versions (with PHP 8.2 minimum) I guess.
Benchmark issues, I don't get it too…

phpstan-baseline.neon
```diff
@@ -126,7 +126,7 @@ parameters:
 			path: src/Components/Expression.php
 
 		-
-			message: "#^Parameter \\#2 \\$pieces of function implode expects array, array\\<string\\>\\|string given\\.$#"
+			message: "#^Parameter \\#2 \\$array of function implode expects array\\|null, array\\<string\\>\\|string given\\.$#"
```
kamil-tekiela, line 129: You need to execute PHPStan with PHP 7.* executable.

src/Tools/ContextGenerator.php
```diff
@@ -153,23 +163,28 @@ public static function sortWords(array &$arr)
...
-     * @return array<int, array<int, array<int, string>>>
+     * @return array<int, list<string>>
      */
     public static function readWords(array $files)
     {
+        /** @psalm-var list<string> $words */
         $words = [];
-        foreach ($files as $file) {
-            $words = array_merge($words, file($file, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES));
+        foreach (array_map('file', $files) as $wordsByFile) {
```
kamil-tekiela, line 174: What exactly is the reason for this change?

src/Tools/ContextGenerator.php
```diff
@@ -325,8 +310,8 @@ public static function formatName($name)
 
         $version = array_map('intval', str_split($versionString, 2));
         /* Remove trailing zero */
-        if ($version[count($version) - 1] === 0) {
-            $version = array_slice($version, 0, count($version) - 1);
+        if (end($version) === 0) {
```
kamil-tekiela, line 313: Personally I never use cursor modifiers such as end(). array_key_last() would be better.

src/Tools/ContextGenerator.php
niconoe-: `FILE_IGNORE_NEW_LINES` and `FILE_SKIP_EMPTY_LINES` are pointless due to this snippet below:
```php
            $value = trim($words[$i]);
            if ($value === '') {
                continue;
            }
```
So I removed them first. 

Than I wanted to get rid of doing `array_merge` in a loop because that's too greedy, so I originally moved to 
```php
$words = array_merge(...array_map(file(...), $files));
```
But array unpacking is not available in PHP 7.2, so when I cherry-picked my changes, I just adapted this part of the code.
Maybe keeping it with a TODO for the next version will be ok?

src/Tools/ContextGenerator.php
```diff
@@ -302,7 +287,7 @@ public static function formatName($name)
     {
         /* Split name and version */
         $parts = [];
-        if (preg_match('/([^[0-9]*)([0-9]*)/', $name, $parts) === false) {
+        if (preg_match('/^(\D+)(\d+)$/', $name, $parts) === 0) {
```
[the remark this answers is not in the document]
niconoe-: It wasn't part of the objectives of the change but PHP Infection noticed it to me, and I needed to clean this up to pass the checks.

`preg_match`  returns either 0 or 1, but only `false` on failure. We don't expect any PCRE failure here, but either a match or not a match. That's why it must be 0 instead of `false`.

src/Tools/ContextGenerator.php
niconoe-: Agreed, but array_key_last is from PHP 7.3.0, and this branch still allows PHP 7.2.0 :cry: 

Using `count(…) - 1` is too greedy and a little bit harder to read IMO than "give me the **end** of the array". And about moving the cursor, as `array_slice` ignores it, it has no effect at all :man_shrugging:.
Do you still prefer the original version?

kamil-tekiela: How about separating the changes into 2 PRs. The context changes go here and the style changes go on master where you can use the newest features.

niconoe-: > How about separating the changes into 2 PRs. The context changes go here and the style changes go on master where you can use the newest features.

Yup, I'll do that. I'll do it also for the `array_map(file(...))` stuff above.

niconoe-: Done :+1:

src/Tools/ContextGenerator.php
```diff
@@ -220,73 +222,56 @@ public static function readWords(array $files)
...
-            $ret .= "\n";
-        }
+    /**
+     * Convert a numeric value representing a set of const to a textual const value.
+     *
+     * @param int $type The numeric value.
+     *
+     * @return string The text to write considering the given numeric value.
+     */
+    private static function numTypeToConst(int $type): string
```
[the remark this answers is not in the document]
niconoe-: Done :+1:

src/Tools/ContextGenerator.php
niconoe-: Back to original state, except for the 2 flags `FILE_IGNORE_NEW_LINES` and `FILE_SKIP_EMPTY_LINES` that I want to remove because they are useless (PHP Infection noticed that).

kamil-tekiela (approved):

williamdes (approved):

niconoe-:
@williamdes (and maybe @MauricioFauth ) any chance to see this integrated soon? Or do you need something from me before? Thanks :slightly_smiling_face:

williamdes:
> @williamdes (and maybe @MauricioFauth ) any chance to see this integrated soon? Or do you need something from me before? Thanks 🙂

For me this is all good, only awaiting a review from @MauricioFauth

PR #571, opened 2024-08-07
Allow using ::class keyword to load a context.

niconoe-:
I noticed that you cannot use the `::class` keyword on contexts to load them.

I updated the Context loaded to manage it, and upgraded the unit tests in charge.

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -599,8 +599,13 @@ abstract class Context
         }
 
         if ($context[0] !== '\\') {
-            // Short context name (must be formatted into class name).
-            $context = self::$contextPrefix . $context;
+            // Could be the fully qualified class name was given, like `ContextDBMS::class`.
+            if (class_exists('\\' . $context)) {
+                $context = '\\' . $context;
+            } else {
+                // Short context name (must be formatted into class name).
+                $context = self::$contextPrefix . $context;
+            }
         }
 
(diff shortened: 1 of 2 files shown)


kamil-tekiela:
I am not at the computer now so I can't check but where and why do we use short class names? Is it just in the tests?

src/Context.php
```diff
@@ -599,8 +599,13 @@ public static function load($context = '')
         }
 
         if ($context[0] !== '\\') {
-            // Short context name (must be formatted into class name).
-            $context = self::$contextPrefix . $context;
+            // Could be the fully qualified class name was given, like `ContextDBMS::class`.
+            if (class_exists('\\' . $context)) {
```
kamil-tekiela, line 603: Are the backslashes necessary?

niconoe-: Unfortunately, yes. The engine requires it  but `::class` does not give it.

niconoe-:
> I am not at the computer now so I can't check but where and why do we use short class names? Is it just in the tests?

That's how the Context loader was built originally: to load the MariaDB 10.7 context for instance, you had to call `Context::load('MariaDb100700');`.
From external projects using SQL-Parser, it's non-intuitive and using the `Context…::class` both enhance the DX but also ensures that you are using an existing context.

williamdes (approved):

PR #577, opened 2024-08-26
Fix #434 - No error when using `UPDATE ... SET`

### LINKED ISSUE #434: No error when using UPDATE ... SET
`UPDATE test SET WHERE` doesn't show any error.

niconoe-:
Fixes #434 

@williamdes / @iifawzi / @MauricioFauth / @liviuconcioiu 

This should also be fixed in 5.10.x and master. How do you want me to proceed to perform this?

### DIFF
--- a/src/Statements/UpdateStatement.php
+++ b/src/Statements/UpdateStatement.php
@@ -145,7 +145,8 @@ class UpdateStatement extends Statement
      * In the update statement, this is used to check that at least one assignment has been set to throw an error if a
      * query like `UPDATE acme SET WHERE 1;` is parsed.
      *
-     * @throws ParserException
+     * @throws ParserException throws the exception, if strict mode is enabled.
+     * @return void
      */
     public function after(Parser $parser, TokensList $list, Token $token)
     {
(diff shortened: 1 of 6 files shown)


MauricioFauth:
Could you please target 5.10.x instead of 5.9.x? Thanks!

niconoe-:
> Could you please target 5.10.x instead of 5.9.x? Thanks!

Done, but the bug will remain en 5.9.x after the merge…

Or do you plan to release a 5.10.0 soon ? :innocent: :yum:

williamdes (approved):

niconoe-:
I updated the fix because it caused failures in SET operations out of an update statement context (like `SET NAMES …`).

Now, that's ok :smiley: . (except a linter in PHP 7.4 which is out of the scope of this PR :shrug: ).

src/Statements/UpdateStatement.php
```diff
@@ -135,4 +139,23 @@ class UpdateStatement extends Statement
      * @var JoinKeyword[]|null
      */
     public $join;
+
+    /**
+     * Function called after the token was processed.
+     * In the update statement, this is used to check that at least one assignment has been set to throw an error if a
+     * query like `UPDATE acme SET WHERE 1;` is parsed.
+     *
+     * @throws ParserException
```
williamdes, line 148: really ?

niconoe-: It actually can throws the ParserException if strict-mode is enabled. This is due to `$parser->error` doing it exactly this way.

I'll add this information in order to make it clearer.

MauricioFauth (approved):

PR #579, opened 2024-09-03
Fix #578: lexer wrongly interprets ".e[0-9]" as a number with scientific notation.

### LINKED ISSUE #578: Unable to parse table names that start with "e1"
Not sure if this is a bug of just a table name with reserved keywords. If the name of my table starts with "e1" the parser is unable to recognize it. It works with any other table name that doesn't start with "e1". Here is my code (SqlParser v.5.10, php version 8.2.16 CLI):

```php
use PhpMyAdmin\SqlParser\Parser         as Parser;

$query = "LOAD DATA LOCAL INFILE '/home/user/myloadfile.csv' 
          IGNORE INTO TABLE erp.e1_table
          FIELDS TERMINATED BY '\t'
          LINES TERMINATED BY '\n'
          IGNORE 0 LINES;";

$parser                                 = new Parser($query);
echo "\n file:   ".$parser->statements[0]->file_name->file;
echo "\n schema: ".$parser->statements[0]->table->database;
echo "\n table:  ".$parser->statements[0]->table->table;
```

This is the actual output:

```
 file:   /home/user/myloadfile.csv
 schema:
 table:  erp
```

Expected output:
```
 file:   /home/user/myloadfile.csv
 schema: erp
 table:  e1_table
```

niconoe-:
@williamdes Here is my fix about #578.

On the first and fourth (because I forgot to move from state 10 to state 4) commits, you can see that token ".e4" is parsed as is and wrongly considered as a number, causing the effect of not being able to detect the "." in "database_name.table_name" as an operator, and cascading to a failure in the parser.

On the second commit, with the fix, you can see ".e4" is no longer a token, but instead "." is a token, and "e4" is another one, which is not a number.

On the third commit… eh, please don't look, I just forgot to remove a debuging stuff 😓

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -671,9 +671,11 @@
       <code>$this-&gt;last</code>
       <code>$this-&gt;last</code>
     </LoopInvalidation>
-    <MixedArrayAccess occurrences="41">
+    <MixedArrayAccess occurrences="43">
       <code>$this-&gt;str[$this-&gt;last + 1]</code>
       <code>$this-&gt;str[$this-&gt;last++]</code>
+      <code>$this-&gt;str[$this-&gt;last]</code>
+      <code>$this-&gt;str[$this-&gt;last]</code>
       <code>$this-&gt;str[$this-&gt;last]</code>
       <code>$this-&gt;str[$this-&gt;last]</code>
       <code>$this-&gt;str[$this-&gt;last]</code>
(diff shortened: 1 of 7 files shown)


MauricioFauth (changes requested):
Could you please add a test with the example from the issue?

niconoe-:
> Could you please add a test with the example from the issue?

Yup, that's done :slightly_smiling_face: 

You can see on `parseLoad8.out` that the tokens are correctly set up, and the statement is having the expected properties without any error.

MauricioFauth (approved):

PR #584, opened 2024-09-20
Refactor getForeignKeys()

kamil-tekiela:
Changed from static to instance and introduced a DTO to help with type inference in PMA. I named it `ForeignKeyData` to avoid clash with `ForeignKey` in PMA.

### DIFF
--- a/src/Utils/ForeignKeyData.php
+++ b/src/Utils/ForeignKeyData.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\SqlParser\Utils;
+
+final class ForeignKeyData
+{
+    /**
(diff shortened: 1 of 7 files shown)


src/Utils/ForeignKeyData.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\SqlParser\Utils;
+
+final class ForeignKeyData
```
MauricioFauth, line 7:
```suggestion
final class ForeignKey
```

I prefer only ForeignKey, as it's way easier to change a class name in PMA than in this lib.

MauricioFauth (changes requested):

williamdes (approved):

PR #590, opened 2024-10-15
Fix keyword parsing

kamil-tekiela:
I am not sure if this is the right base and I don't know what the psalm and phpstan changes are about. 

This should fix https://github.com/phpmyadmin/phpmyadmin/issues/19320


williamdes (approved):
This looks okay, thank you

MauricioFauth (approved):

PR #598, opened 2024-11-10
Remove dependency on zumba/json-serializer

MauricioFauth:
Instead of deserializing the data and then comparing it with the actual objects, this serializes the actual object and then compares it with the serialized data.

Adds duplicated information and the type of objects are lost, but this allows for the removal of zumba/json-serializer dependency.

We can develop a specialized serializer for this or use another method to test this data later on.


tests/data/bugs/fuzz2.out
```diff
@@ -1,15 +1,21 @@
...
-        "@type": "PhpMyAdmin\\SqlParser\\Lexer",
+        "KEYWORD_NAME_INDICATORS": [
+            "FROM",
+            "SET",
+            "WHERE"
+        ],
+        "OPERATOR_NAME_INDICATORS": [
+            ",",
+            "."
+        ],
```
williamdes, line 12: can you implement something to drop the constants ?
they add quite some data and will cause all files to have to be updated on a new value added

tests/data/bugs/fuzz2.out
```diff
@@ -1,15 +1,21 @@
...
         "str": "WITH](",
         "len": 6,
         "last": 6,
         "list": {
-            "@type": "PhpMyAdmin\\SqlParser\\TokensList",
```
williamdes, line 9: having the type was not too bad, but we can drop it

williamdes:
Well, if we can trim down a bit the diff by adding back `@types` and removing the consts that would be okay to me
Why do you want to drop this lib, is there a reason ?

MauricioFauth:
> Why do you want to drop this lib, is there a reason ?

Nothing against this lib. It's because we don't actually need a full serializer for theses tests. I think we don't even need to use json for it.

MauricioFauth:
I just noticed that removing the `PhpMyAdmin\SqlParser\Tools\CustomJsonSerializer` is a BC break and this PR can't be merged. Because the class is in the `src` directory and it's not marked with the `@internal` annotation.
I'll annotate the classes inside `PhpMyAdmin\SqlParser\Tools` with `@internal` and do this changes in the `master` branch.

PR #599, opened 2024-11-15
Add Translator::setLocale() method

MauricioFauth:
The Translator::setLocale() takes precedence over $GLOBALS['lang']. This makes possible to deprecate the usage of the $GLOBALS['lang'].

### DIFF
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -5,7 +5,10 @@ declare(strict_types=1);
 namespace PhpMyAdmin\SqlParser;
 
 use PhpMyAdmin\MoTranslator\Loader;
+use PhpMyAdmin\MoTranslator\Translator as MoTranslator;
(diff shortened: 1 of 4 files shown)


src/Translator.php
```diff
@@ -16,32 +18,40 @@ class Translator
...
+    private static $locale = '';
+
     /**
      * Loads translator.
      *
      * @return void
      */
     public static function load()
     {
+        if (! class_exists(Loader::class)) {
```
kamil-tekiela, line 42: Why is the guard needed now?

MauricioFauth: Just because this method is public. But it should be private.
Maybe throwing an exception is better. Or annotating it as internal method, or both.

williamdes (approved):

PR #611, opened 2025-01-30
Fix #610 - Add the missing options to handle in the context of ALTER TABLE statement: "UNIQUE" and "UNIQUE KEY"

### LINKED ISSUE #610: `ALTER TABLE ADD UNIQUE (KEY)` is not recognized as an alter option
When trying to parse some SQL like:
```sql
ALTER TABLE testtable 
ADD UNIQUE KEY name_of_the_unique_constraint (`UNIQUE_COLUMN`);
```
or 
```sql
ALTER TABLE testtable 
ADD UNIQUE name_of_the_unique_constraint (`UNIQUE_COLUMN`);
```

the parser considers every token after the `ADD` keyword as "unknown", while adding indexes or constraint works well. This is due to the lack of the presence of the options "UNIQUE" and "UNIQUE KEY" in the `AlterOperation::$TABLE_OPTIONS` array.

I'll provide a PR about it, fixing also the unit tests about the parser expecting unknown tokens rather than well accepted options.

niconoe-:
Fix #610 by adding the missing options to handle in the context of ALTER TABLE statement: "UNIQUE" and "UNIQUE KEY".

### DIFF
--- a/tests/data/parser/parseAlterTableAddUniqueKey1.out
+++ b/tests/data/parser/parseAlterTableAddUniqueKey1.out
@@ -217,24 +217,22 @@
                         "options": {
                             "@type": "PhpMyAdmin\\SqlParser\\Components\\OptionsArray",
                             "options": {
-                                "1": "ADD"
+                                "1": "ADD",
+                                "2": "UNIQUE KEY"
                             }
                         },
-                        "field": null,
+                        "field": {
+                            "@type": "PhpMyAdmin\\SqlParser\\Components\\Expression",
+                            "database": null,
+                            "table": null,
+                            "column": "name_of_the_unique_constraint",
+                            "expr": "name_of_the_unique_constraint",
+                            "alias": null,
+                            "function": null,
+                            "subquery": null
+                        },
                         "partitions": null,
                         "unknown": [
-                            {
-                                "@type": "@10"
-                            },
-                            {
-                                "@type": "@11"
-                            },
-                            {
-                                "@type": "@12"
-                            },
-                            {
-                                "@type": "@13"
-                            },
                             {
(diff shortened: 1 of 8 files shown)


tests/data/parser/parseAlterTableAddUniqueKey1.out
```diff
@@ -0,0 +1,266 @@
...
+                        "unknown": [
+                            {
+                                "@type": "@14"
+                            },
+                            {
+                                "@type": "@15"
+                            },
+                            {
+                                "@type": "@16"
+                            }
```
williamdes, line 244: is it difficult to have the last tokens understood by the parser ?

niconoe-: To be honest, I don't know. I just have an issue with a project of my own that uses the parser which fails to recognise "ADD UNIQUE" and "ADD UNIQUE KEY" options.

AFAIK, it's the same with "ADD INDEX" or "ADD FOREIGN KEY", so maybe the recognition of further unknown tokens should be part of another PR, dedicated to this.

niconoe-:
I fixed this in the 5.11.x. Would it be possible to have it on 5.10.x too? If yes, should I provide another PR from the 5.10.x branch?

tests/data/parser/parseAlterTableAddUniqueKey1.out
williamdes: Sure

MauricioFauth (approved):

MauricioFauth:
Merged by f40c620af9bdefd6eae97c21e934d17e835c05d5.

PR #613, opened 2025-02-03
Add handling of index hints on join clauses too. Fixes #593 and #497.

### LINKED ISSUE #497: FORCE INDEX is not parsed correctly
This first query results in the following errors when linting, while the second one has no problems
```
#1: Unexpected token. (near "=" at position 58)
#2: Unexpected token. (near "uno" at position 60)
#3: Unexpected token. (near "." at position 63)
#4: Unexpected token. (near "id" at position 64)
```

```sql
SELECT *
FROM uno
JOIN dos FORCE INDEX (two_id) ON dos.id = uno.id
```

```sql
SELECT *
FROM uno
JOIN dos  ON dos.id = uno.id
```

### LINKED ISSUE #593: Building statement with FORCE INDEX results in wrong SQL
Hello,

I have query that's not parsed or built properly when `FORCE INDEX` is used:
```sql
SELECT a.id, a.name, b.order_id, b.total
FROM customers a
INNER JOIN orders b FORCE INDEX (idx_customer_id)
    ON a.id = b.customer_id
WHERE a.status = 'active'
```

The query becomes:

```sql
SELECT a.id, a.name, b.order_id, b.total
FROM customers AS `a` FORCE INDEX (idx_customer_id)
INNER JOIN orders AS `b`
WHERE a.status = 'active'
```

- `FORCE INDEX` got moved from table `orders` onto `customers`.
- The `a.id = b.customer_id` join condition was removed.

I see the same linting errors as #497, so this is probably related.

<details>
  <summary>Example code</summary>

```php
<?php
require __DIR__ . '/vendor/autoload.php';

$in = <<<SQL
SELECT a.id, a.name, b.order_id, b.total
FROM customers a
INNER JOIN orders b FORCE INDEX (idx_customer_id)
    ON a.id = b.customer_id
WHERE a.status = 'active'
SQL;

$parser = new PhpMyAdmin\SqlParser\Parser($in);
$statement = $parser->statements[0];
echo $statement->build();
```
</details>

niconoe-:
Fixes #593 
Fixes #497 

This is to be merged in the `master` branch. After taking another look, it does not seem to be a BC Break as I thought in #593, as the indexHints should just be added from the JOIN clause handling, not being removed from the SELECT statement.

I'll do another PR for the 5.10.x and 5.11.x versions, as this one uses new features from v6 the v5 doesn't have yet.

### DIFF
--- a/src/Components/JoinKeyword.php
+++ b/src/Components/JoinKeyword.php
@@ -6,6 +6,7 @@ namespace PhpMyAdmin\SqlParser\Components;
 
 use PhpMyAdmin\SqlParser\Component;
 use PhpMyAdmin\SqlParser\Parsers\Conditions;
+use PhpMyAdmin\SqlParser\Parsers\IndexHints;
 
 use function array_search;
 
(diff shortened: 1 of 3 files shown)


niconoe-:
AFAIK, all mutants here are actually from previous states, so I'm not sure it's the role of this PR to try to kill em. What do you think about @williamdes ?

williamdes:
> AFAIK, all mutants here are actually from previous states, so I'm not sure it's the role of this PR to try to kill em. What do you think about @williamdes ?

yeah okay, no worries then. Thank you for checking

MauricioFauth:
Closing this in favor of #614.

PR #616, opened 2025-02-09
Remove nullability from IndexHint

kamil-tekiela:
This confused me but I checked the documentation and it turns out that neither $type nor $indexOrKey can be null. https://dev.mysql.com/doc/refman/8.4/en/index-hints.html

### DIFF
--- a/tests/data/parser/parseSelectIndexHintErr1.out
+++ b/tests/data/parser/parseSelectIndexHintErr1.out
@@ -252,7 +252,7 @@
                     {
                         "@type": "PhpMyAdmin\\SqlParser\\Components\\IndexHint",
                         "type": "FORCE",
-                        "indexOrKey": null,
+                        "indexOrKey": "",
                         "for": null,
                         "indexes": [
                             {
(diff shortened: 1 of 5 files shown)


tests/data/parser/parseSelectIndexHintErr1.out
```diff
@@ -252,7 +252,7 @@
                     {
                         "@type": "PhpMyAdmin\\SqlParser\\Components\\IndexHint",
                         "type": "FORCE",
-                        "indexOrKey": null,
+                        "indexOrKey": "",
```
williamdes, line 255: I guess that's a bit less intuitive for developers

williamdes (approved):
But all okay

tests/data/parser/parseSelectIndexHintErr1.out
kamil-tekiela: How so? Either way, it's an invalid value. The developer should check the parser errors.

williamdes: Oh okay, it's empty but the parser would have errors

MauricioFauth (approved):

PR #618, opened 2025-02-18
Remove `/tests` folder from pull

bert-w:
The `/tests` folder is several MBs big. It sounds like developers don't need this folder when they depend on this package and install it through Composer, so I have excluded it in `.gitattributes`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,7 +2,7 @@
 *.out linguist-language=json
 
 /tools export-ignore
-/tests/benchmarks export-ignore
+/tests export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 .editorconfig export-ignore


MauricioFauth:
On the other hand, finding a way to reduce the size of the tests would be great!

williamdes:
@MauricioFauth I think we can re open and accept it
On Debian I will switch to git mode and get the files

bert-w:
@MauricioFauth what are the use cases exactly? I don't know in which other ways this repository is being used atm.

MauricioFauth:
@remicollet Could you please comment on this? In [2017](https://github.com/phpmyadmin/sql-parser/commit/5e9ed9878aa9335c32597e42f88bfce256075bf6#commitcomment-20562869), you were against adding the `tests` directory to the `.gitattributes` file with the attribute `export-ignore`. Has that changed since then? If not, nothing will change. Thank you.

williamdes:
> @MauricioFauth what are the use cases exactly? I don't know in which other ways this repository is being used atm.

Ref: https://tracker.debian.org/pkg/phpmyadmin-sql-parser

Debian imports the source code of projects into it's own git (I can work without git)
To (import/detect versions) of sql-parser Debian uses [uscan](https://manpages.debian.org/unstable/devscripts/uscan.1.en.html), config: https://salsa.debian.org/phpmyadmin-team/sql-parser/-/blob/debian/5.10.3-1_bpo12+1/debian/watch?ref_type=tags#L5
Uscan is currently configured to pull the vendor tarball
But Debian also ensures that it's packages work, so it "needs" to run the upstream tests.
That's the reason why the folder is required.
That said, there is a git mode like in this example: https://salsa.debian.org/phpmyadmin-team/tcpdf/-/blob/debian/6.8.2+dfsg-1_bpo12+1/debian/watch?ref_type=tags#L3

Switching to git mode looks like that: https://salsa.debian.org/phpmyadmin-team/tcpdf/-/commit/4b43e784532c7ddf5b75342969fc7e24931f269f
I had to do it because TCPDF removed the examples/ folder recently
And examples are useful and ran as tests.

PS: the uscan config (debian/watch) is also used to detect new upstream releases to package.

I will let @remicollet explain to us how it works for them/Fedora

remicollet:
> I will let @remicollet explain to us how it works for them/Fedora

In Fedora we use a git snapshot to workaround this broken sources issue

I still think that using .gitattributes is a shame. It is a terrible workaround for the lack of an installer for PHP libraries.
But I gave up the fight, so do the fuck you want

williamdes:
@MauricioFauth let's get it clean for users ?
We have much more users now than the project is used for Larastan

MauricioFauth:
Let's extract the end-to-end tests and ignore them (as they're the big ones), so we can keep the unit tests.

Issue #615, opened 2025-02-06
row_number() over sql parse error


attson:
### eg: 

```php
        $sql = <<<EOF
select id,
       pid,
       appid,
       appname,
       row_number() over (partition by pid, appid) as `group_row_number`
from game group by appname
EOF;

        $parser = new \PhpMyAdmin\SqlParser\Parser($sql);

        dump($parser->statements[0]->build());
        dump($parser->statements[0]->expr);
```

output:

```sql
SELECT id, pid, appid, appname, row_number()  (partition by pid, appid) AS `over` FROM game GROUP BY appname

...

  4 => PhpMyAdmin\SqlParser\Components\Expression
    +database: null
    +table: null
    +column: "row_number"
    +expr: "row_number()  (partition by pid, appid)"
    +alias: "over"
    +function: "row_number"
    +subquery: null
  }

...
```

### expected

return an incorrect alias, expected to be group_row_number.

### env
```
PHP 8.1.30 (cli) (built: Sep 27 2024 04:07:29) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.30, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.30, Copyright (c), by Zend Technologies
    with Xdebug v3.3.2, Copyright (c) 2002-2024, by Derick Rethans

phpmyadmin/sql-parser 5.10.1
```

phpmyadmin/sql-parser 5.10.3  same error

williamdes:
I think this may be a duplicate of #188 or be related to it

PR #620, opened 2025-03-18
Add MariaDB Contexts 11.8 and 12.0.

niconoe-:
Sources : 
- https://mariadb.com/kb/en/what-is-mariadb-118/
- https://mariadb.com/kb/en/changes-improvements-in-mariadb-12-0/

PR #623, opened 2025-04-09
Fix Window function handling that failed because of "OVER" keyword case-sensitive comparison.

### LINKED ISSUE #188: Support for window functions like RANK
The validator marks queries with an alias (with or without `AS`) after [window functions](https://mariadb.com/kb/en/library/window-functions/) like `RANK`, `DENSE_RANK`, `ROW_NUMBER`, `CUME_DIST`, etc. as incorrect:
```SELECT RANK() OVER(ORDER BY NULL) rank```

```
[
  {
    "message": "An alias was previously found. (near <code>rank</code>)",
    "fromLine": 0,
    "fromColumn": 34,
    "toLine": 0,
    "toColumn": 38,
    "severity": "error"
  },
  {
    "message": "Unexpected token. (near <code>rank</code>)",
    "fromLine": 0,
    "fromColumn": 34,
    "toLine": 0,
    "toColumn": 38,
    "severity": "error"
  }
]
```

Support for this type of functions is probably not present yet, but weirdly enough the validator doesn't complain for queries without alias.

PHPMyAdmin 4.7.9
MariaDB 10.3.1 (window functions are supported since 10.2.0)

Update: these will also be supported in [MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html)

niconoe-:
Fixes #188 and #615

### DIFF
--- a/src/Components/Expression.php
+++ b/src/Components/Expression.php
@@ -15,6 +15,7 @@ use function implode;
 use function is_array;
 use function rtrim;
 use function strlen;
+use function strtoupper;
 use function trim;
 
 /**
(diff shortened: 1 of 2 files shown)


attson:
Good job! when will this version be released

niconoe-:
Hello @williamdes ! Any news about integrating this and releasing it?

I'm also expecting #620 to be released.

Thanks a lot!

MauricioFauth (approved):
Thank you!

PR #635, opened 2025-11-14
UUID column type

### LINKED ISSUE #634: uuid datatype not recognized for MariaDB 10.7+
Since MariaDB 10.7 a column could have a [datatype uuid](https://mariadb.com/docs/server/reference/data-types/string-data-types/uuid-data-type), but this type is not recognized by the parser:

```sql
CREATE TABLE `users` (
  `id` uuid NOT NULL PRIMARY
);
```

Image

MoonE:
Fixes #634


williamdes (approved):
Thank you, that looks good

PR #646, opened 2026-03-04
Refactoring formatter

kamil-tekiela:
These PR adds DTO for formatting options. In this repo, it seems that only type is ever provided. In the main repo only 

> $this->response->addJSON(['sql' => Formatter::format($query, ['line_ending' => ' ', 'indentation' => ''])]);

This PR removes the ability to provide custom formats. I don't know if any other project relies on it and I can add it back, but I removed it to make the code simpler. I think it would be better to keep that option even if it's not used in PMA, but please let me know which is better. 

The options `parts_newline` and `indent_parts` were removed as I could not see them being used anywhere.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1753,15 +1753,15 @@ parameters:
 			path: src/Utils/CLI.php
 
 		-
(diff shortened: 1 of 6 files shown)


williamdes (approved):
Does this create a breaking change?

MauricioFauth (approved):
Thank you!

PR #654, opened 2026-04-22
ELSE expressions terminate early when encountering an END token

SharpDerek:
CASE statements with an ELSE returning a value with the "tablename.columnname" format fail due to encountering an END token and parsing it as if it were a column instead. This fixes the issue by checking to see if the current token is an END and breaks before column processing.

### DIFF
--- a/src/Components/Expression.php
+++ b/src/Components/Expression.php
@@ -275,8 +275,10 @@ class Expression extends Component
 
                     $isExpr = true;
                 } elseif (
-                    $brackets === 0 && strlen((string) $ret->expr) > 0 && ! $alias
-                    && ($ret->table === null || $ret->table === '')
+                    $token->value === "END" || (
+                        $brackets === 0 && strlen((string) $ret->expr) > 0 && ! $alias
+                        && ($ret->table === null || $ret->table === '')
+                    )
                 ) {
                     /* End of expression */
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Thank you for this fix. If possible please reset your branch to 5.11.x and cherry pick your work and force push.
It will allow us to have the fix on 5.11.x
Else just let me know and I will do it manually

williamdes:
Also, please add test cases

SharpDerek:
> Also, please add test cases

Here's the most succinct test case I can provide:
```
SELECT
    CASE
        WHEN table_a.column_a = 'FOO'
        THEN table_b.column_a
        ELSE table_b.column_b
END AS foo
```

Output before change:
```
SELECT
    CASE foo
    ELSE table_b.column_b AS `END`
END
```

Output after change:
```
SELECT
    CASE WHEN table_a.column_a = 'FOO'
    THEN table_b.column_a
    ELSE table_b.column_b
END AS `foo`
```

SharpDerek:
If this gets merged into 5.11.x, would it be put into 6.0.0 as well? That is the version that we are currently running and can't really downpatch.

SharpDerek:
@williamdes Is there anything more you need from me on this?

PR #656, opened 2026-04-24
Fix #655 - Missing space before UNION when building CREATE VIEW

### LINKED ISSUE #655: Missing space before UNION when building CREATE VIEW
## Summary

When `CreateStatement::build()` reconstructs a `CREATE VIEW ... AS SELECT ... UNION ...`
statement, the output misses the whitespace between the built `SELECT` and the
trailing body tokens. Result: invalid SQL such as `... WHERE 3 = 3union all (...)`.

This was previously reported downstream in phpMyAdmin:
- phpmyadmin/phpmyadmin#19692
- phpmyadmin/phpmyadmin#18515 (same bug, older)

## Reproduction

```php
use PhpMyAdmin\SqlParser\Parser;

$sql = "CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`localhost` SQL SECURITY DEFINER "
     . "VIEW `v1` AS select 1 AS `a` where 3 = 3 union all (select 2 AS `a`)";

echo (new Parser($sql))->statements[0]->build();
```

Actual output:
```
CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`localhost` SQL SECURITY DEFINER VIEW `v1`  AS SELECT 1 AS `a` WHERE 3 = 3union all (select 2 AS `a`)
```
(note: `3 = 3union all` — no space).

Expected: a space before `union all`, producing valid SQL.

## Environment

- Library version: current `master` (head `f440917e`)
- PHP: 8.4

predictor2718:
Fixes #655.

`CreateStatement::build()` glued the built `SELECT` directly to the trailing body
tokens when rebuilding a `CREATE VIEW ... AS SELECT ... UNION ...`, producing
invalid SQL like `... WHERE 3 = 3union all (...)`.

Insert a single space between `$builtStatement` and the body tokens only when
both sides are non-empty and neither boundary character is whitespace, so the
paren-wrapped LHS case (`AS (select ...) union all (...)`) doesn't get a double
space.

Includes a regression test in `tests/Builder/CreateStatementTest.php`.

Also addresses phpMyAdmin reports:
- phpmyadmin/phpmyadmin#19692
- phpmyadmin/phpmyadmin#18515

Signed-off-by: predictor2718 <245527909+predictor2718@users.noreply.github.com>

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -2102,7 +2102,7 @@ parameters:
 
 		-
 			message: "#^Dynamic call to static method PHPUnit\\\\Framework\\\\Assert\\:\\:assertEquals\\(\\)\\.$#"
-			count: 34
+			count: 37
 			path: tests/Builder/CreateStatementTest.php
 
 		-
--- a/src/Statements/CreateStatement.php
+++ b/src/Statements/CreateStatement.php
@@ -16,7 +16,9 @@ use PhpMyAdmin\SqlParser\Statement;
 use PhpMyAdmin\SqlParser\Token;
 use PhpMyAdmin\SqlParser\TokensList;
 
+use function ctype_space;
 use function is_array;
+use function substr;
 use function trim;
 
 /**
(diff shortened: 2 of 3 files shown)


williamdes (approved):
Thanks!

PR #454, opened 2023-05-31
Remove Misc class

kamil-tekiela:
Follow up from #451 but this time I move the class to `SelectStatement` where it belongs due to feature envy.

PR #658, opened 2026-06-21
Avoid using a null alias as an array offset in getAliases()

ardittirana:
Resolves the `phpmyadmin/sql-parser` side of phpmyadmin/phpmyadmin#20313.

### Problem
`SelectStatement::getAliases()` builds a table-alias map:

```php
$tables[$thisDb][$expr->alias] = $expr->table;
```

for every `FROM`/`JOIN` expression. When a table has no alias, `$expr->alias` is `null`, so this uses **null as an array offset**, which is deprecated as of PHP 8.5:

> Using null as an array offset is deprecated, use an empty string instead

phpMyAdmin surfaces this as an `ErrorException` (`SelectStatement.php:614`) during table export on PHP 8.5.

### Fix
Skip expressions without an alias, mirroring the existing `isset($expr->alias) && $expr->alias !== ''` guards a few lines above. The null-keyed entry was never read anyway — the only consumer looks up `$tables[$thisDb][$expr->table]`, whose offset is always a non-empty string — so `getAliases()` output is unchanged.

### Notes
- Removed the now-resolved `PossiblyNullArrayOffset` entry from `psalm-baseline.xml` (verified via `psalm --update-baseline`; only that entry changed).
- The sibling report #20312 points at `Utils/Misc.php`, which no longer exists — `getAliases()` was moved into `SelectStatement` in #454 — so this covers the live code path.

### Tests
- Added a `getAliasesProvider` case mixing an unaliased table with an aliased one (the null-alias path), asserting unchanged output.
- Full suite green: 1020 tests, 10244 assertions. `phpcs`/`psalm` clean for the change.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1359,9 +1359,6 @@
     <MixedArrayOffset occurrences="1">
       <code>$tables[$thisDb][$expr-&gt;table]</code>
     </MixedArrayOffset>
-    <PossiblyNullArrayOffset occurrences="1">
-      <code>$tables[$thisDb]</code>
-    </PossiblyNullArrayOffset>
   </file>
   <file src="src/Utils/Query.php">
     <InvalidNullableReturnType occurrences="1">
--- a/src/Utils/Misc.php
+++ b/src/Utils/Misc.php
@@ -68,7 +68,7 @@ class Misc
                 ];
             }
 
-            if (null === $expr->alias || $expr->alias === '') {
+            if ($expr->alias === null || $expr->alias === '') {
                 continue;
             }
 
--- a/src/Utils/Misc.php
+++ b/src/Utils/Misc.php
@@ -68,6 +68,10 @@ class Misc
                 ];
             }
 
+            if (null === $expr->alias || $expr->alias === '') {
+                continue;
+            }
+
             if (! isset($tables[$thisDb])) {
                 $tables[$thisDb] = [];
             }
--- a/tests/Utils/MiscTest.php
+++ b/tests/Utils/MiscTest.php
@@ -132,6 +132,25 @@ class MiscTest extends TestCase
                     ],
                 ],
             ],
+            [
+                'SELECT mytable.a x, o.b y FROM mytable JOIN other o ON o.id = mytable.id',
+                'shop',
+                [
+                    'shop' => [
+                        'alias' => null,
+                        'tables' => [
+                            'mytable' => [
+                                'alias' => null,
+                                'columns' => ['a' => 'x'],
+                            ],
+                            'other' => [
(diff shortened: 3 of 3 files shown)


src/Statements/SelectStatement.php
```diff
@@ -607,6 +607,13 @@ public function getAliases(string $database): array
                 ];
             }
 
+            // Only record an alias => table mapping when an alias exists.
+            // Using a null alias as an array offset is deprecated as of
+            // PHP 8.5, and such an entry is never read back below anyway.
+            if (! isset($expr->alias) || ($expr->alias === '')) {
```
williamdes: It is better to check explicitly null

src/Statements/SelectStatement.php
```diff
@@ -607,6 +607,13 @@ public function getAliases(string $database): array
                 ];
             }
 
+            // Only record an alias => table mapping when an alias exists.
+            // Using a null alias as an array offset is deprecated as of
```
williamdes: The comment lines look like they where added by an AI
Please remove this line and below it adds no value

tests/Builder/StatementTest.php
```diff
@@ -155,6 +155,28 @@ public static function getAliasesProvider(): array
                     ],
                 ],
             ],
+            [
+                // An unaliased table alongside an aliased one. The unaliased
+                // table must not break alias resolution; previously its null
+                // alias was used as an array offset (deprecated in PHP 8.5).
```
williamdes: Remove this one

williamdes:
Thank you for this contribution!
if you used AI please make it add a Co author line.
Also I would like that this PR branch is based on 5.11.x so next version has your fix and not just 6.0

And finally please fix lint step

ardittirana:
Thanks for the review @williamdes, and for the helpful pointers. I've updated the PR:

- **Rebased onto `5.11.x`** and changed the PR base accordingly, so the fix lands in the next 5.11 release and merges up to 6.0. On this branch the affected code lives in `src/Utils/Misc.php` (`getAliases()`), which is also the spot reported in #20312, so this covers that path too.
- **Explicit null check**: the guard is now `if (null === $expr->alias || $expr->alias === '')`.
- **Removed the noisy comments** in both the source and the test — agreed they didn't add value.
- The change keeps `getAliases()` output identical (the null-keyed entry was never read back; the only consumer indexes `$tables[$thisDb][$expr->table]`), and I removed the now-resolved `PossiblyNullArrayOffset` entry from the Psalm baseline.
- A regression test is added to `MiscTest::getAliasesProvider`.

Full suite is green locally. Please let me know if you'd like anything else adjusted.

williamdes (approved):
Thanks !

